PR #13, opened 2020-04-02
Minor fixes to Tests and investigation on #6

### LINKED ISSUE #6: consistency in returned types
It's not clear when we should take:
```
pytorch_model(torch.from_numpy(X))[0]  #  RandomForestClassifier
```
vs. say
```
pytorch_model(torch.from_numpy(X))[1]   # DecisionTreeClassifier
```

In the end, we want it to be something simple like what we compare to such as
```
model.predict(X)
```

interesaaat:
Fixes #6.


tests/test_lightgbm_converters.py
```diff
@@ -1,7 +1,8 @@
 """
-Tests sklearn TruncatedSVD converter
+Tests Skleran LightGBM converters.
```
ksaur: fix the "skleran" -> "sklearn"

requirements.txt
```diff
@@ -1,6 +1,6 @@
 numpy>=1.15
 scikit-learn==0.21.3
-torch==1.3.1
+torch>=1.3.1
```
ksaur: let's fix to 1.4.0 so that it works with python 3.8

tests/test_lightgbm_converters.py
```diff
@@ -60,25 +75,21 @@ def _run_lgbm_regressor_converter(self, num_classes, extra_config={}):
...
-
-    # multi
-    def test_lgbm_multi_regressor_converter(self):
-        self._run_lgbm_regressor_converter(3)
+        self._run_lgbm_regressor_converter(1000)
 
-    # gemm
+    # gemm regressor
     def test_lgbm_gemm_regressor_converter(self):
-        self._run_lgbm_regressor_converter(3, extra_config={"tree_implementation": "gemm"})
```
ksaur, line 73: just curious, how much longer do these take to run?

tests/test_sklearn_random_forest_converters.py
```diff
@@ -27,26 +27,38 @@ def _run_random_forest_classifier_converter(self, num_classes, extra_config={},
...
+    # perf_tree_trav classifier
+    def test_random_forest_perf_tree_trav_classifier_converter(self):
+        self._run_random_forest_classifier_converter(2, extra_config={"tree_implementation": "perf_tree_trav"})
+
     # multi classifier
     def test_random_forest_classifier_multi_converter(self):
         self._run_random_forest_classifier_converter(3)
 
-    # gemm classifier
+    # gemm multi classifier
```
ksaur, line 53: should we also make these 1000 to make the other file?

tests/test_lightgbm_converters.py
interesaaat: fixed.

tests/test_lightgbm_converters.py
interesaaat: The value here does not matter for the runtime. It's not like binary vs multiclass. In that case it matters :)

tests/test_sklearn_random_forest_converters.py
interesaaat: No no this is ok. 1000 was to get wider margins to the regression.

PR #18, opened 2020-04-06
Ksaur/notebooks

ksaur:
Notebooks for LGBM and XGB with random data


notebooks/LGBM-randomdata.ipynb
```diff
@@ -0,0 +1,122 @@
...
+      ]
+     },
+     "execution_count": 2,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# Create and train a model (LightGBM in this case)\n",
+    "model = lgb.LGBMClassifier()\n",
```
ksaur, line 43: @interesaaat - are there other times we should show here for comparison?

ksaur: (see http://13.77.169.87:8889/notebooks/notebooks/LGBM-randomdata.ipynb)

notebooks/XGB-randomdata.ipynb
```diff
@@ -0,0 +1,130 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "12.4 ms ± 26.3 µs per loop (mean ± std. dev. of 3 runs, 100 loops each)\n"
```
ksaur: see also http://13.77.169.87:8889/notebooks/notebooks/XGB-randomdata.ipynb

interesaaat (commented):
Two things:
* should we remove the current example in the notebook loading a pre-trained model? I think these are better, no?
* we will eventually have to change the API if the other PR is merged. Perhaps we can then show an example for XGBoostClassifier and XGBoostRegressor since the API is different. WDYT?

ksaur:
* Ok, on the pre-trained model...i think that it does add something because it is legit data and seems a bit more real.  but i don't feel strongly.

* on changing the API, ok, i'll make sure to update it after yours merges! and good idea on showing a regressor; the thought crossed my mind too

PR #19, opened 2020-04-06
API review for convert_sklearn

interesaaat:
* Remove initial_types from the converter API
* Rename Skl2PyTorch container into HBPyTorch

### DIFF
--- a/tests/test_xgboost_converters.py
+++ b/tests/test_xgboost_converters.py
@@ -8,7 +8,6 @@ import numpy as np
 import torch
 import xgboost as xgb
 from hummingbird import convert_sklearn
-from onnxconverter_common.data_types import FloatTensorType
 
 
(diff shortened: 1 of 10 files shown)


tests/test_xgboost_converters.py
```diff
@@ -69,7 +68,8 @@ def _run_xgb_regressor_converter(self, num_classes, extra_config={}):
             y = np.random.randint(num_classes, size=100)
 
             model.fit(X, y)
-            pytorch_model = convert_sklearn(model, [("input", FloatTensorType([1, 200]))], extra_config=extra_config)
+            extra_config["n_features"] = 200
```
interesaaat: This is the only place where the API actually gets worst. I wasn't able to find a way to find the number of features from the XGBoost regressor.

ksaur (approved):

interesaaat:
Fix #20.

PR #32, opened 2020-04-26
Add few tests to check that we actually pick the correct tree implementation

interesaaat:


tests/test_sklearn_decision_tree_converters.py
```diff
@@ -30,44 +48,44 @@ def _run_decision_tree_classifier_converter(self, num_classes, extra_config={},
             )
 
     # Binary classifier
-    def test_decision_tree_classifier_binary_converter(self):
-        self._run_decision_tree_classifier_converter(2)
+    def test_random_forest_classifier_binary_converter(self):
```
ksaur, line 51: Why did you decide to change this back? Also, should we change back the file name so the tests names and filename match?

ksaur (approved):

tests/test_sklearn_decision_tree_converters.py
interesaaat: Because the test is actually running random forest. I think that the file name is ok since random forest is a bunch of decision trees. And we do have decision trees experiments later in the file.

ksaur: ooookie dokie.

PR #41, opened 2020-04-29
Kasaur/setup feedback

ksaur:
Are there other things you can think of to add to Troubleshooting at the moment?

Also, does this link to `virtenv` work?  (using [scikit](https://scikit-learn.org/stable/install.html) as example)


interesaaat:
I don't think of any. We add more as we go.

Issue #14, opened 2020-04-03
Add predict and predict_proba methods to Skl2PyTorch class


interesaaat:

interesaaat:
I did an initial investigation on this. So the implementation requires to just implement `predict` and `predict_proba` in `Skl2PyTorch`. The problems are the following:
* It should work for both classification and regression. For regression we need to check that `predict_proba` is not used, and `predict` will return the full array for regression while only `[0]` for classification. This is sort of hard without  flagging into the Skl2PyTorch wrapper rather we are doing regression or classification.
* This will only work for PyTorch and not for the other DNN runtimes. We can probably make it work for TorchScript and TVM too, but for ONNX it will be hard because you don't do `model(data)` for inference but you actually need to create a ORT session. Bottom line, for ONNX having just a model is not enough.

PR #48, opened 2020-05-04
This PR adds a tighter API on tops of the converter one

interesaaat:
This API should be shorter and nicer to use. I think I fixed already all docs, tests and notebooks.


README.md
```diff
@@ -12,7 +12,7 @@ Currently we support [these](https://github.com/microsoft/hummingbird/blob/devel
 
 ## Installation
 
-This was tested on Python 3.7 on a Linux machine.  It is recommended to use a virtual environment (See: [python3 venv doc](https://docs.python.org/3/tutorial/venv.html) or [Using Python environments in VS Code](https://code.visualstudio.com/docs/python/environments).)
+Hummingbird was tests on Python 3.6. and 3.7 on Linux, Windows and MacOS machines.  It is recommended to use a virtual environment (See: [python3 venv doc](https://docs.python.org/3/tutorial/venv.html) or [Using Python environments in VS Code](https://code.visualstudio.com/docs/python/environments).)
```
ksaur: s/testd/tested/g

README.md
```diff
@@ -28,13 +28,13 @@ See also [Troubleshooting](TROUBLESHOOTING.md) for common problems.
 
 See the [notebooks](notebooks) section for examples that demonstrate use and speedups.
 
-In general, the syntax is very similar to [skl2onnx](https://github.com/onnx/sklearn-onnx), as hummingbird started as a fork of that project.
+In general, Hummingbird syntax is very intuitive and minimal. Basically, to run your traditional ML model on DNN frameworks, you only need to `import hummingbird` and add `to('dnn_framework')` to your code. Below an example using a [LightGBM](https://lightgbm.readthedocs.io/en/latest/) model and [PyTorch](https://pytorch.org/) as target framework.
```
ksaur: I think drop "Basically," and just have "To run your..."

also: "Below *is* an example using"

hummingbird/supported.py
```diff
@@ -0,0 +1,110 @@
...
+from lightgbm import LGBMClassifier, LGBMRegressor
+
+
+def _build_sklearn_operator_list():
+    """
+    Put all suported Sklearn operators on a list.
+    """
+    if sklearn_installed():
+        # Tree-based models
+        from sklearn.ensemble import (
```
ksaur, line 38: Hmm...now we don't have one "clean" list anywhere.  And we link to this list in several places. Do you have suggestions?

ksaur: Ex: I simply dropped the list from https://github.com/microsoft/hummingbird/pull/49

README.md
interesaaat: Fixed.

README.md
interesaaat: Fixed

hummingbird/supported.py
interesaaat: Yea. Do you think it is too much expecting users to read the supported file and find the operators? Perhaps I can list in the comment in supporter ops and backend. In this way it will also show in the documentation. WDYT?

ksaur: Yah reading through the file would be asking a lot, especially as the code grows around it.  I think we need a nice list somewhere.  A comment at the top could work. Or we could maintain one on the wiki...as long as we remember to maintain it ;)

ksaur:
You have 3 of the notebooks, but not [this one](https://github.com/microsoft/hummingbird/blob/master/notebooks/blog_example.ipynb)

hummingbird/supported.py
interesaaat: Agree that the problem is maintenance. What do you prefer? Eventually we can start with both and then ditch one of the two as we grow.

ksaur: I think  I like the wiki approach.I'll do that now.

interesaaat:
Closes #14.

PR #63, opened 2020-05-11
dockerfile to build and launch this in a notebook

ksaur:
I expect the link to the dockerfile will work when it is in the master branch

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,7 +38,7 @@ git clone https://github.com/microsoft/hummingbird.git .
 pip install -e .
 ```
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -38,6 +38,12 @@ git clone https://github.com/microsoft/hummingbird.git .
 pip install -e .
 ```
 
+We also provide a simple [Dockerfile](https://github.com/microsoft/hummingbird/blob/master/Dockerfile) that you can customize to your preferred development environment. This Dockerfile will create a Jupyter notebook environment and also install development tools such as `flake8` and `coverage`.
```
ksaur: Do you think this should be here, or on the main readme under install instructions?

Issue #61, opened 2020-05-11
Add support for Visual Studio Codespaces


markusweimer:
One way to resolve #60 is to add explicit support for dev containers to the Hummingbird repo. The documentation for that is [here](https://docs.microsoft.com/en-us/visualstudio/online/reference/configuring). Specifically, we could add a Dockerfile and settings for VS Code to use it. I am happy to give that a try if someone has a Dockerfile I can base that on.

ksaur:
I cannot access Visual Studio Codespaces.  But here is https://github.com/microsoft/hummingbird/blob/kasaur/docker/Dockerfile

markusweimer:
Thanks! I have fiddled around a bit and made this:

https://github.com/markusweimer/hummingbird/tree/devcontainer/.devcontainer

Is there interest to have this as a PR?

ksaur:
Yes, that would be great, thank you!

ksaur:
Closing via PR #71

interesaaat:
Can we have a documentation on how to use this thing in CONTRIBUTING.md?

ksaur:
writing now.

PR #71, opened 2020-05-12
Added devcontainer configuration.

markusweimer:
This adds a new folder `.devcontainer` which contains the config necessary for the Visual Studio Code dev container setup. This is especially useful on VS Codespaces (and maybe Github Codespaces). 

The setup is very bare bones, but is enough to follow the instructions in the contribution's guide. We can improve on the comfort of the setup if we find this whole thing useful :)


markusweimer:
This addresses #61 directly, and #61 indirectly.

Issue #72, opened 2020-05-13
Make the dev container more comfy


markusweimer:
The current dev container is derived off of a super barebones image. Not even `less` is installed. I will be looking into finding a more comfy base image, probably starting with the ones found [here](https://github.com/microsoft/vscode-dev-containers).

markusweimer:
I have something working [here](https://github.com/markusweimer/hummingbird/tree/mweimer/72). It uses a devcontainer base image with better tools installed, and also configures VSCode to use `bash` instead of `sh`.

There are very many additional niceties that could be done to the user setup (better prompt and such). However, I think that is better left to the per-user dotfiles repo.

This container is still using Python 3.6, while the one for jupyter uses Python 3.7. Which version does Hummingbird actually require? I took 3.6 from `setup.py`.

markusweimer:
@interesaaat, can you help with this?

> This container is still using Python 3.6, while the one for jupyter uses Python 3.7. Which version does Hummingbird actually require? I took 3.6 from `setup.py`.

interesaaat:
Sorry I forgot about replying here. I think 3.6 is safer but we support python 3 in general.

markusweimer:
Thanks, I will send the change as a PR then

ksaur:
closing re: #75

PR #75, opened 2020-05-14
Updated the dockerfile to be more comfy.

markusweimer:
This changes the Dockerfile to use a devcontainer base image.

Also, this adds a setting to default to `bash` instead of `sh`.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,7 +1,23 @@
-# basic setup
-FROM oryxprod/python-3.6
-RUN apt-get update && apt-get -y update
-RUN pip -q install pip --upgrade
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
 
-# Install dev dependencies
-RUN pip -q install flake8 coverage autopep8 pre-commit pylint
+FROM mcr.microsoft.com/vscode/devcontainers/python:0-3.6
+
+#
+# Update the OS and maybe install packages
+#
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update \
+   && apt-get upgrade -y  \
+   && apt-get -y install --no-install-recommends build-essential \
+   && apt-get autoremove -y \
+   && apt-get clean -y \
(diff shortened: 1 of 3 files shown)


markusweimer:
This addresses #72

interesaaat:
How close are we with having a single docker file?

markusweimer:
> How close are we with having a single docker file?

I believe the two docker images serve different purposes. The one in `.devcontainer` is for users of VS Code. The one in `.` is for people using the notebooks. Maybe we could base one on the other? Maybe in some hierarchy where the bare bones base image is for the build server, the slightly nicer one for VS Code and the last one is for Notebook users?

interesaaat:
Yes make sense. So should be stick for all images with a version of python? If we are going with 3.6, can you please update the beafy docker image? @ksaur (for visibility)

markusweimer:
> So should be stick for all images with a version of python?

That makes sense, I think. At least for development of hummingbird, sticking to one Python version avoids accidentally moving the requirements to a newer Python version.

ksaur:
I don't think a single docker image is desired at this time because they are two very different purposes.   (I spent hours trying to unify the two files earlier this week, and we discussed offline and decided it was a bad idea.)    Also, since no one really uses codespaces yet, it's really hard to say how it will evolve at this time.

yes sticking to 3.6 python for now.

ksaur:
And the main Dockerfile is also for people doing development in any editor _of their choice_ which may include vscode, not just notebooks.  I don't want to be locked into any particular editor, especially a web-based one.

markusweimer:
> And the main Dockerfile is also for people doing development in any editor of their choice which may include vscode, not just notebooks. I don't want to be locked into any particular editor, especially a web-based one.

Agreed. However, I think that is true for the container definintion in `.devcontainer` as well. There is nothing in the dockerfile which is specific to VS Code. The VS Code specific stuff is in `devcontainer.json` in that folder.

markusweimer:
Shall we update `./Dockerfile` in this PR to use Python 3.6 or make a separate PR for that?

ksaur:
Sure, yes please.  Just do %s/3.7/3.6/g on `./Dockerfile`. Thanks!

I just tested it offline and verified it works.  (i guess some automated testing on `Dockerfile`s would be a nice-to-have, but it would slow down things, so it would have to be outside the regular github actions workflow.)

markusweimer:
Done. `./Dockerfile` now also references Python 3.6

PR #83, opened 2020-05-29
Update README.md

thvasilo:
Not sure if the example ocde or the text is the one that should be fixed here, just taking a stab.


interesaaat:
Thanks for catching this Theodore!

Issue #11, opened 2020-04-02
Add GradientBoostingRegressor


interesaaat:

bfgray3:
I'd enjoy working on this feature if nobody has started on it yet.

interesaaat:
Go for it!

NicolasHug:
`GradientBoostingRegressor` has an init argument https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.GradientBoostingRegressor.html that specifies the first estimator that is used to initialize the predictions, before the first GB iteration.

I think treelite raises an error when the `init` argument is different form `zero`. I haven't checked with them directly but I assume it's because they can only compile trees, not arbitrary estimators.

Would this be an issue with hummingbird as well?

Regarding https://github.com/microsoft/hummingbird/issues/64: the `HistGradientBoosting` estimators don't accept an `init` parameter, but they do have an internal `_baseline_prediction` which usually corresponds to the average of the target and which is automatically added to any prediction. This thing *isn't* a tree and it isn't part of the `_predictors` list, it's just a private float attribute (EDIT: in multiclass settings it's an array of size `n_classes`). So maybe  https://github.com/microsoft/hummingbird/issues/64  will need to take care of that manually?

Alternatively I think we could consider converting `_baseline_prediction` into a single-node tree in sklearn (I don't think we can do much internally about `init` though :/)

PR #87, opened 2020-05-30
[WIP] Add sklearn's HistGradientBoosting

### LINKED ISSUE #64: Add sklearn's HistGradientBoosting

ahmedkrmn:
Closes #64:

- This PR adds sklearn's [HistGradientBoostingClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingClassifier.html) to hummingbird.

- It modifies the two functions `convert_sklearn_gbdt_classifier()` and `get_parameters_for_sklearn_common()` to handle the HistGradientBoostingClassifier model attributes.

- It also updates the documentation to include the new HistGradientBoostingClassifier.


ahmedkrmn:
@interesaaat getting the value of Alpha is what's left.

@NicolasHug I checked your [comment](https://github.com/microsoft/hummingbird/issues/11#issuecomment-636325188) which references using the `_baseline_prediction` attribute of the HistGradientBoostingClassifier model for doing this.
I noticed that it switches between being a `<class 'numpy.ndarray'>` and `<class 'numpy.float64'>` during different iterations, so could you elaborate on that?

NicolasHug:
Right my bad, it's not always a float. In multiclass problems it's a numpy array of size `n_classes` because we build `n_classes` trees at each iteration. The decision function for a given class corresponds to the sum of all the predictions of the trees of that class, and the predicted probability to belong to a class is a softmax of these sums. It is the same idea as in multinomial logistic regression, where we get `n_classes` weight vectors.

Since we build `n_classes` set of trees, we need to compute one initial prediction per class.

ahmedkrmn:
@NicolasHug OK great, thanks a lot for this info!

@interesaaat I'm still not sure about how we are going to map the `_baseline_prediction` to alpha and also how `init_` is being mapped to alpha is still vague for me. For example, why do we have to take the natural logarithm here?
```python
alpha = [[np.log(operator.raw_operator.init_.class_prior_[i]) for i in range(n_classes)]]
```

hummingbird/ml/operator_converters/_tree_commons.py
```diff
@@ -169,20 +169,42 @@ def get_tree_params_and_type(tree_infos, get_tree_parameters, extra_config):
...
+        # Use -1 instead of 0 for nodes with no rights
+        rights = [idx if idx != 0 else -1 for idx in rights]
+
+        features = [trees.nodes[x]["feature_idx"] for x in range(len(trees.nodes))]
+        thresholds = [trees.nodes[x]["threshold"] for x in range(len(trees.nodes))]
+
+        values = [trees.nodes[x]["value"] for x in range(len(trees.nodes))]
+        # Convert to numpy.ndarray with shape (s, 1, 1)
+        values = np.array(values)
+        values = values.reshape((values.shape[0], 1, 1))
```
interesaaat: Not sure why you need to this. Isn't reshape`(-1, 1)` enough?

hummingbird/ml/supported.py
```diff
@@ -34,11 +35,15 @@ def _build_sklearn_operator_list():
     Put all suported Sklearn operators on a list.
     """
     if sklearn_installed():
+        # Enable experimental to import HistGradientBoostingClassifier
+        from sklearn.experimental import enable_hist_gradient_boosting
```
interesaaat, line 42: Why is this needed?

interesaaat (commented):
Looks good. I haven't test it yet tho. I left a couple of comments. We will need to fix the coverage, but let's finish the coding first.

interesaaat:
> @NicolasHug OK great, thanks a lot for this info!
> 
> @interesaaat I'm still not sure about how we are going to map the `_baseline_prediction` to alpha and also how `init_` is being mapped to alpha is still vague for me. For example, why do we have to take the natural logarithm here?
> 
> ```python
> alpha = [[np.log(operator.raw_operator.init_.class_prior_[i]) for i in range(n_classes)]]
> ```

I believe that something like `alpha = np.array([operator.raw_operator._baseline_prediction.flatten().tolist()])` should work. The `init_` part is not clear to me either. @NicolasHug is the `init_` here used as in GradientBoosting? In that case we can just do a check as treelite (as pointed by Nicolas [here](https://github.com/microsoft/hummingbird/issues/11#issuecomment-636325188)) and open as issue if the value is different than `zero`. In this way we can have the operator in and figure out later this corner case.

hummingbird/ml/supported.py
NicolasHug: it's because the HistGBDT estimators are still experimental and are not subject to the 2-versions deprecation cycles, so we requires users to explicitly say "I want to use this feature anyway"

When the estimators become stable, this will have no effect

NicolasHug:
**Short version**

> I believe that something like alpha = np.array([operator.raw_operator._baseline_prediction.flatten().tolist()]) should work

Yes that should work (modulo shape conversion). As far as I understand, reading the existing code in `convert_sklearn_gbdt_classifier`, `ALPHA` corresponds to the initial prediction of the boosting process, which is precisely `_baseline_prediciton`.


**Long version**

> is the init_ here used as in GradientBoosting?

it is used in the GradientBoostingXXX estimators but not in the HistGradientBoostingXXX estimator. 

In the GradientBoostingXXX estimators, we allow users to pass an `init` parameter which is an arbitrary estimator object that specifies the very first prediction, i.e. the value at which predictions are initialized when starting the boosting process. (Kinda like the value you would use to initialize weights when doing gradient descent)

In the HistGBDT code, there is no `init` parameter and we just hardcode it. The specific value isn't super important, but it corresponds to the mean of the target in regression (with LS loss), and to the log-odds ratio in binary classification, because that's the "best"  we can do w.r.t. the loss while only looking at the targets.

Predicting the log-odds ratio is also the default when `init=None` in the GradientBoostingXXX estimators, though in this case a `dummy` estimator is used. That dummy estimator always predicts a constant prediction which is the prior of the positive class, stored as its `class_prior_[1]` attribute.

The current code for the GradientBoostingClassifier is correct: when `init=None`, `ALPHA` is set to the log-odds ratio to replicate the behavior of the estimator 
(nice job to whoever wrote this BTW, this isn't obvious!)

```py
    # Get the value for Alpha.
    if operator.raw_operator.init == "zero":
        alpha = [[0.0]]
    elif operator.raw_operator.init is None:
        if n_classes == 1:
            alpha = [[np.log(operator.raw_operator.init_.class_prior_[1] / (1 - operator.raw_operator.init_.class_prior_[1]))]]
        else:
            alpha = [[np.log(operator.raw_operator.init_.class_prior_[i]) for i in range(n_classes)]]
    else:
        raise RuntimeError("Custom initializers for GBDT are not yet supported in Hummingbird."
```

This is for the GradientBoostingClassifier. For the HistGradientBoosting estimators, there is no dummy-estimator, and the log-odds ratio is just in `_baseline_prediction`. Basically `_baseline_prediction` corresponds to `np.log(operator.raw_operator.init_.class_prior_[1] / (1 - operator.raw_operator.init_.class_prior_[1])`, so you just have to replace that

interesaaat:
> The current code for the GradientBoostingClassifier is correct: when `init=None`, `ALPHA` is set to the log-odds ratio to replicate the behavior of the estimator
> (nice job to whoever wrote this BTW, this isn't obvious!)

@scnakandala is the genius behind everything :) 

Thanks for the detailed explanation @NicolasHug! Perhaps renaming `Alpha` to `base_prediction` or `base_score` (as in XGBoost) will be a bit more clear. WDYT?

hummingbird/ml/operator_converters/_tree_commons.py
ahmedkrmn: `values` has a shape of `(x, 1, 1)` in case of HistGradientBoosting.
Doing `values.reshape(-1, 1)` we will have a shape of `(x, 1)` instead of `(x, 1, 1)`. If this won't cause any problems I can modify it.

interesaaat: Yes, can you please put them as `-1, 1`?

interesaaat (changes requested):
Looks good! 2 asks:
- Can you please add in this PR the tests Karla wrote for you?
- Can you please merge the changes we pushed into master?

Will merge after these are fixed (plus the reshape).

ahmedkrmn:
Hi @interesaaat, 
After merging master, I found that commit 0f8232879b90bce is causing failing tests for `test_sklearn_gbdt_converters.py`.

ahmedkrmn:
Hi @interesaaat,
I've added the test by @ksaur. Unfortunately, it fails on 9 out of the 11 tests.
I'll push the changes so that you can have a look on the failing output.

interesaaat:
> Hi @interesaaat,
> I've added the test by @ksaur. Unfortunately, it fails on 9 out of the 11 tests.
> I'll push the changes so that you can have a look on the failing output.

I see. Apparently the probabilities are off for some reason. Were the tests passing on your side before commit  0f82328?

ahmedkrmn:
> > Hi @interesaaat,
> > I've added the test by @ksaur. Unfortunately, it fails on 9 out of the 11 tests.
> > I'll push the changes so that you can have a look on the failing output.
> 
> I see. Apparently the probabilities are off for some reason. Were the tests passing on your side before commit [0f82328](https://github.com/microsoft/hummingbird/commit/0f8232879b90bce1d4aa77b6df89d7b741d92d39)?

Yeah I tried before merging, they failed as well.

tests/test_sklearn_histgbdt_converters.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class TestSklearnHistGradientBoostingClassifier(unittest.TestCase):
+    # Check tree implementation
+    def test_gbdt_implementation(self):
+        warnings.filterwarnings("ignore")
+        X = np.random.rand(10, 1)
+        X = np.array(X, dtype=np.float32)
+        # y = np.random.randint(2, size=10)
+
```
KranthiGV: nit: Did we miss the actual test here? Currently, it doesn't test any. Can we please add any assertions or remove this method?

KranthiGV:
> > > Hi @interesaaat,
> > > I've added the test by @ksaur. Unfortunately, it fails on 9 out of the 11 tests.
> > > I'll push the changes so that you can have a look on the failing output.
> > 
> > 
> > I see. Apparently the probabilities are off for some reason. Were the tests passing on your side before commit [0f82328](https://github.com/microsoft/hummingbird/commit/0f8232879b90bce1d4aa77b6df89d7b741d92d39)?
> 
> Yeah I tried before merging, they failed as well.

Looking at the test results, the predicted probability is off by a high value (many a times even the first decimal digit is different while the expected decimal precision in test is 6). That's why all the 9 tests that assert on predicted probabilities fail.

Did they ever pass locally for you @ahmedkrmn? We can backtrack using that info. Also, is the predicted class correct?

Also, interesaaat / ksaur, how do you people generally debug these issues?

interesaaat:
> Also, interesaaat / ksaur, how do you people generally debug these issues?

Well, we start with regression, 1 tree and see why the results don't match. Then we do regression multiple trees -> binary -> multiclass. In the worst case we check the partial results within Sklearn execution and where they start to differ wrt Hummingbird.

I can take a look at this.

ahmedkrmn:
Please let me know if there is anything I can do to help.

ahmedkrmn:
@interesaaat any new updates?

interesaaat:
No sorry I have been working in the ONNX PR. I am done with it so I will look into this next.

interesaaat:
Hey @ahmedkrmn  I fixed the problem with the conversion. Basically, learning rate and normalization for `hist_gbdt` are already computed upfront and we don't need to do them on our side. 

2 questions:
- can you please merge with master (sorry)
- can you strip the `hist_gbdt` implementation from the `gbdt` one? At the beginning they were quite similar, but now they differ quite a bit so I prefer to have separate implementations. You can even just add the code in `_tree_commons.py` into `gbdt.py` at this point.

ahmedkrmn:
> Hey @ahmedkrmn I fixed the problem with the conversion. Basically, learning rate and normalization for `hist_gbdt` are already computed upfront and we don't need to do them on our side.
> 
> 2 questions:
> 
> * can you please merge with master (sorry)
> * can you strip the `hist_gbdt` implementation from the `gbdt` one? At the beginning they were quite similar, but now they differ quite a bit so I prefer to have separate implementations. You can even just add the code in `_tree_commons.py` into `gbdt.py` at this point.

Great Matteo, thanks for the update!
I'll use a separate `convert_sklearn_hist_gbdt_classifier` function for dealing with histGBDT.

Regarding moving the code of `_tree_commons.py` to `gbdt.py`, I can see that `get_parameters_for_sklearn_common()` function is used in`decision_tree.py`, so is moving it a good idea?
Or what you mean is that we create a separate `get_parameters_for_sklearn_common()` for histGBDT and put it in `gbdt.py` and leave the original function as is in the `_tree_commons.py`?

interesaaat:
> > Hey @ahmedkrmn I fixed the problem with the conversion. Basically, learning rate and normalization for `hist_gbdt` are already computed upfront and we don't need to do them on our side.
> > 2 questions:
> > 
> > * can you please merge with master (sorry)
> > * can you strip the `hist_gbdt` implementation from the `gbdt` one? At the beginning they were quite similar, but now they differ quite a bit so I prefer to have separate implementations. You can even just add the code in `_tree_commons.py` into `gbdt.py` at this point.
> 
> Great Matteo, thanks for the update!
> I'll use a separate `convert_sklearn_hist_gbdt_classifier` function for dealing with histGBDT.
> 
> Regarding moving the code of `_tree_commons.py` to `gbdt.py`, I can see that `get_parameters_for_sklearn_common()` function is used in`decision_tree.py`, so is moving it a good idea?
> Or what you mean is that we create a separate `get_parameters_for_sklearn_common()` for histGBDT and put it in `gbdt.py` and leave the original function as is in the `_tree_commons.py`?

Sorry, I only meant the code related to `hist_gbdt`m like from like 183 in `gbdt`. The other side of the condition is ok if stays in `_tree_commons.py`

ahmedkrmn:
Hi @interesaaat,
Does `test_sklearn_histgbdt_converters.py` pass on your local machine after your additions?
It's still failing on mine.

interesaaat:
> Hi @interesaaat,
> Does `test_sklearn_histgbdt_converters.py` pass on your local machine after your additions?
> It's still failing on mine.

Yes it is passing all tests in my machine. Are you getting the same failures or something different? Like, all tests are still failing?

hummingbird/ml/operator_converters/_tree_commons.py
```diff
@@ -180,22 +180,11 @@ def get_parameters_for_sklearn_common(tree_infos):
...
-
-        values = [trees.nodes[x]["value"] for x in range(len(trees.nodes))]
-        # Convert to numpy.ndarray with shape (s, 1, 1)
-        values = np.array(values)
-        values = values.reshape((values.shape[0], 1, 1))
+        features = [n[2] if n[2]!=0 else 0 for n in trees.nodes]
+        thresholds = [n[3] if n[3]!=0 else -1 for n in trees.nodes]
+        lefts = [n[4] if n[4]!=0 else -1 for n in trees.nodes]
+        rights = [n[5] if n[5]!=0 else -1 for n in trees.nodes]
+        values = [[n[0]] if n[0]!=0 else [-1] for n in trees.nodes]
```
ahmedkrmn, line 187: The tests are giving errors this time instead of failures. The reason is the `_find_max_depth()` in `_tree_commons.py`.
When I reverted these lines back to using `lefts = [trees.nodes[x]["left"] for x in range(len(trees.nodes))]` ... etc, all 10 tests passed.
I will inspect this now.

ahmedkrmn: @interesaaat I found the issue!
The problem with accessing tree nodes by index is that it's not guaranteed that this index will always be equivalent to the property that you're trying to fetch.
Using the string name of the property for fetching is before guarantees the correct result.

hummingbird/ml/operator_converters/_tree_commons.py
```diff
@@ -180,22 +180,11 @@ def get_parameters_for_sklearn_common(tree_infos):
...
-        thresholds = [trees.nodes[x]["threshold"] for x in range(len(trees.nodes))]
-
-        values = [trees.nodes[x]["value"] for x in range(len(trees.nodes))]
-        # Convert to numpy.ndarray with shape (s, 1, 1)
-        values = np.array(values)
-        values = values.reshape((values.shape[0], 1, 1))
+        features = [n[2] if n[2]!=0 else 0 for n in trees.nodes]
+        thresholds = [n[3] if n[3]!=0 else -1 for n in trees.nodes]
+        lefts = [n[4] if n[4]!=0 else -1 for n in trees.nodes]
+        rights = [n[5] if n[5]!=0 else -1 for n in trees.nodes]
```
ahmedkrmn: Hi @interesaaat , the test is failing because you've used wrong indices for accessing `lefts` and `rights`.
The index of `lefts` is `5` and for `rights` it's `6`.
I suggest accessing by the string name of the property instead of the index to avoid such bugs.

ahmedkrmn:
@interesaaat When running the tests locally, sometimes one test fails as in the case with this last failing build. If you rerun the tests again, all of them pass, so that's a strange behavior!

hummingbird/ml/operator_converters/_tree_commons.py
interesaaat: Agreed! I think that we probably have different sklearn versions, and they changed the indexing between versions.

interesaaat:
Can you please point me to the failing case? Is it the multiclass? This is probably due to rounding errors and we can fix it momentarily by increasing the tolerance.

ahmedkrmn:
> Can you please point me to the failing case? Is it the multiclass? This is probably due to rounding errors and we can fix it momentarily by increasing the tolerance.

I ran the tests 20 times. It's always either that they all pass or 9/10 pass.

**Summary:**
10/20 all passed.
02/20 failed at `test_HistGBDT_classifier_binary_converter`.
02/20 failed at `test_HistGBDT_perf_tree_trav_multi_classifier_converter`.
02/20 failed `test_HistGBDT_gemm_classifier_converter`.
01/20 failed at `test_HistGBDT_tree_trav_multi_classifier_converter`.
01/20 failed at `test_HistGBDT_shifted_labels_converter`.

interesaaat:
I can take a look. Can you please rebase with master? After my pass we can then merge this PR!

ahmedkrmn:
> I can take a look. Can you please rebase with master? After my pass we can then merge this PR!

Done!

ahmedkrmn:
Thanks a lot, @interesaaat,
This was really an interesting PR to work on!

PR #94, opened 2020-06-01
Change requirements structure

### LINKED ISSUE #93: Setup better requirements structure
I am thinking about following something along [this line ](https://stackoverflow.com/questions/17803829/how-to-customize-a-requirements-txt-for-multiple-environments)

interesaaat:
Closes #93.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -38,11 +38,11 @@ jobs:
         flake8 . --count  --max-complexity=10 --max-line-length=127 --statistics
     - name: Test with pytest
       run: |
-        pip install -r requirements.txt && pip install -e . && pip install pytest
+        pip install -e . && pip install pytest
         pytest
     - name: Coverage
       run: |
-        pip install -r requirements.txt && pip install -e . && pip install coverage
+        pip install -e . && pip install coverage
         coverage run -m pytest tests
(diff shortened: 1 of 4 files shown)


requirements/test.txt
```diff
@@ -0,0 +1,4 @@
+-r common.txt
```
ksaur, line 1: Ok, so the plan is to have this separate testing file, and then leave regular `requirements.txt` in place as an alternative to `pip install` yes?

.github/workflows/pythonapp.yml
```diff
@@ -28,21 +28,18 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -r requirements/test.txt .
```
ksaur, line 31: does the "." add something here?

ksaur (approved):

requirements/test.txt
ksaur: Ah ok, i see your comments  in #93

.github/workflows/pythonapp.yml
interesaaat: I was surprised too. basically without "." it was not installing hummingbird, but only the requirements.

PR #96, opened 2020-06-02
Add python 3.5, 3.7 and 3.8 to CI pipeline

### LINKED ISSUE #95: Add more python versions to CI pipeline

interesaaat:
Closes #95.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -38,11 +38,11 @@ jobs:
         flake8 . --count  --max-complexity=10 --max-line-length=127 --statistics
     - name: Test with pytest
       run: |
-        pip install -r requirements.txt && pip install -e . && pip install pytest
+        pip install -e . && pip install pytest
         pytest
     - name: Coverage
       run: |
-        pip install -r requirements.txt && pip install -e . && pip install coverage
+        pip install -e . && pip install coverage
         coverage run -m pytest tests
         coverage xml
         MINIMUM=70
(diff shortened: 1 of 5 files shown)


.github/workflows/pythonapp.yml
```diff
@@ -18,31 +18,31 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
```
ksaur, line 23: I'm curious,  how much slower does this make things  run?
(i'm not suggesting we remove it as we need it!)

README.md
```diff
@@ -2,6 +2,7 @@
 
 [![PyPI version](https://badge.fury.io/py/hummingbird-ml.svg)](https://badge.fury.io/py/hummingbird-ml)
 ![](https://github.com/microsoft/hummingbird/workflows/Python%20application/badge.svg?branch=develop)
+![](https://img.shields.io/badge/python-3.5%20%7C%203.6%20%7C%203.7%20%7C%203.8-blue)
```
ksaur, line 5: oh good. MOARBADGES!!!

ksaur (approved):

ksaur:
Also....why does  this make coverage  go  up?? Is there something only  one of the  versions is  hitting??

interesaaat:
> Also....why does this make coverage go up?? Is there something only one of the versions is hitting??

That's a good question! Let me dig into it.

.github/workflows/pythonapp.yml
interesaaat: I don't know, I haven't test it but it should be minimal because they run concurrently. It's a bit more because 3.8 takes more than the others for some reason.

interesaaat:
Ok it is related to the abstract class. Before it was not covered for some reason apparently.

PR #97, opened 2020-06-02
WIP: add mac and win runners to github actions

interesaaat:


setup.py
```diff
@@ -24,6 +22,33 @@
     if start_pos >= 0:
         long_description = long_description[start_pos:]
 
+common_requirements = ["numpy>=1.15", "Cython", "onnxconverter-common>=1.6.0"]
+if sys.platform == "darwin" or sys.platform == "linux":
+    common_requirements.append("torch>=1.4.0")
+else:
+    if sys.version_info >= (3, 5, 0):
```
ksaur, line 29: oh interesting; does this work?

ksaur: maybe we could 
 do something like https://github.com/opencv/openvino_training_extensions/blob/develop/pytorch_toolkit/nncf/setup.py#L70

interesaaat:
Fixed by #128

PR #106, opened 2020-06-04
adding in py 3.5

ksaur:


interesaaat:
Good catch!

KranthiGV:
We should do this in pypi as well. It currently shows that one needs Python 3.6+

interesaaat:
Yes. Will update this once we push 0.0.2. Setup.py is already correctly showing >= 3.5.

KranthiGV:
Ah, okay! Didn't realize that. Thanks..

PR #119, opened 2020-06-10
Removing not required deps from install; Bumping SKL version; adding mac runners

### LINKED ISSUE #3: upgrade and test new sklearn version
we need to upgrade sklearn version.  (currently  scikit-learn==0.21.3).  To do this, we need to accomodate some API changes in the newer version

ex: Imputer is deprecated (was in preprocessing). Now, do :from sklearn.impute import SimpleImputer

### LINKED ISSUE #68: Remove not required dependencies
Should we leave `sklearn`? From one site having `sklearn` will help first time users, but on the other side, everyone taking a dependency on `hummingbird` will have to install `sklearn` even if it is not used.

ksaur:
Working on Issue #68

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -17,7 +17,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: [ubuntu-latest, windows-2019]
     strategy:
       matrix:
         python-version: [3.5, 3.6, 3.7, 3.8]
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -36,6 +36,8 @@ jobs:
     - name: Run basic tests without extra
       run: |
         pytest tests/test_no_extra_install.py
+    - name: Coverage on basic tests without extra
+      run: |
         coverage run -a -m pytest tests/test_no_extra_install.py
     - name: If mac, install libomp to facilitate lgbm install
       if: matrix.os == 'macOS-latest'
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install .[extra,tests,docs]
+        pip install  --no-cache-dir .[extra,tests,docs]
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-2019]
+        os: [ubuntu-latest, macos-latest]
         python-version: [3.5, 3.6, 3.7, 3.8]
 
     steps:
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install hummingbird-ml[extra]
+        pip install .[extra,tests,docs]
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements/test.txt .
+        pip install hummingbird-ml[extra]
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -17,9 +17,10 @@ on:
 jobs:
   build:
 
-    runs-on: [ubuntu-latest, windows-2019]
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ubuntu-latest, windows-2019]
         python-version: [3.5, 3.6, 3.7, 3.8]
 
     steps:
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -28,9 +28,15 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+    - name: Install basic dependencies
       run: |
         python -m pip install --upgrade pip
+        pip install  --no-cache-dir .
+    - name: Run basic tests without extra
+        pytest tests/test_no_extra_install.py
+        coverage run -a -m pytest tests/test_no_extra_install.py
+    - name: Install extra dependencies
+      run: |
         pip install  --no-cache-dir .[extra,tests]
     - name: Lint with flake8
       run: |
@@ -40,10 +46,10 @@ jobs:
         flake8 . --count  --max-complexity=10 --max-line-length=127 --statistics
     - name: Test with pytest
       run: |
-        pytest
+        pytest --ignore=tests/test_no_extra_install.py
     - name: Coverage
       run: |
-        coverage run -m pytest tests
+        coverage run -a -m pytest tests --ignore=tests/test_no_extra_install.py
         coverage xml
         MINIMUM=70
         SCORE=$(coverage report -m | tail -n 1 | awk '{print $NF}' | rev | cut -c2- | rev)
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -31,14 +31,14 @@ jobs:
     - name: Install basic dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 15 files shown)


hummingbird/ml/_utils.py
```diff
@@ -35,6 +35,7 @@ def sklearn_installed():
 
         return True
     except ImportError:
+        warnings.warn("To use sklearn with Hummingbird, please install `sklearn`, or: `pip install hummingbird-ml[extra]`")
```
ksaur: This will cause the exception higher up the stack and print a warning, but maybe we should rework

Dockerfile
```diff
@@ -12,10 +12,10 @@ RUN update-alternatives --set python /usr/bin/python3.6
 RUN pip3 -q install pip --upgrade
 
 # Install HB
-RUN pip3 install hummingbird-ml
+RUN pip3 install hummingbird-ml[docs,tests,extra]
 
-# Install HB dev tools
-RUN pip3 install flake8 coverage autopep8 jupyter pre-commit pytest
+# Install additional dev gools for notebook env
```
interesaaat: goals?

.github/workflows/pythonapp.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements/test.txt .
+        pip install  --no-cache-dir .[extra,tests,docs]
```
interesaaat: no need for docs here. We will need it later I guess.

CONTRIBUTING.md
```diff
@@ -35,7 +35,7 @@ The simplest setup is:
 mkdir hummingbird
 cd hummingbird
 git clone https://github.com/microsoft/hummingbird.git .
-pip install -e .
+pip install -e .[docs,tests,extra]
```
interesaaat, line 38: Here no docs

Dockerfile
ksaur: This dockerfile is sort of bloated and needs to be rewritten to have a separate build phase to not have all of the extra  stuff.  But...as it is is, it already has all of the extra stuff.

I think i'll open a "good first issue" for this.

.github/workflows/pythonapp.yml
ksaur: yah i was kind of thinking ahead to when this is all automated, but no need for it now.

CONTRIBUTING.md
ksaur: if someone is contributing, i want them to have docs i think (hope)

ksaur:
This is a little tricky because we haven't really tested much without having all three (skl, xgb,lgbm) installed.  It "APPEARS" that the tests are passing, but that is because we are only testing the "all installed" case.

I've been testing this manually in a clean container, but I'm going to pause for a second to setup some tests like:

```
    - name: Install basic dependencies
      run: |
        python -m pip install --upgrade pip
        pip install  --no-cache-dir .
    - name: Run basic tests without extra
        # SOMETHING HERE
    - name: Install extra dependencies
      run: |
        pip install  --no-cache-dir .[extra,tests]
```

interesaaat:
Yes now that we have separate dependencies it would be really good if we can run some test without all dependencies.

ksaur:
BTW, it seems XGB has a dependency on SKL.  I started a clean container, only installed XGB (not SKL, not LGBM), and I get:
```
>>> model = xgb.XGBRegressor()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/dist-packages/xgboost/sklearn.py", line 146, in __init__
    raise XGBoostError('sklearn needs to be installed in order to use this module')
xgboost.core.XGBoostError: sklearn needs to be installed in order to use this module
```

ksaur:
And LBGM installs scikit-learn as a dep.
```
root@2c5fb8629da4:~/hummingbird# pip3 freeze | grep sci
scipy==1.4.1
root@2c5fb8629da4:~/hummingbird# pip3 install lightgbm>=2.2
root@2c5fb8629da4:~/hummingbird# pip3 freeze | grep sci
scikit-learn==0.23.1
scipy==1.4.1`
```

ksaur:
But, this new installer will at least allow JUST SKL without the requirement on XGB and LGBM.

@interesaaat - Do you think we should make SKL a required dependency since all 3 of our targets require it? (I believe also the onnx work will require it at least for now?)

interesaaat:
Yes if lgbm requires skl, than also the lgbm converter for onnx will require it, so let's have it! We can remove it later on in case we have some library that does not need skl.

ksaur:
On python 3.8, we're hitting https://github.com/scikit-learn/scikit-learn/issues/15328 because we're using SKL 0.21.3

With the new installer, for some stupid reason even though I have Cython listed before Sklearn, it's installing sklearn first and failing on
```
      File "sklearn/utils/setup.py", line 8, in configuration
        from Cython import Tempita
    ModuleNotFoundError: No module named 'Cython'
```

So i think we should either upgrade to a newer version of SKL (see https://github.com/microsoft/hummingbird/issues/3), or not support python 3.8.  WDYT @interesaaat ?   I believe the newer SKL versions was only a problem with our Imputer code (yet to be added)

.github/workflows/pythonapp.yml
```diff
@@ -28,10 +28,17 @@ jobs:
...
-    - name: Install dependencies
+    - name: Install basic dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements/test.txt .
+        pip install  --no-cache-dir .[tests]
+    - name: Run basic tests without extra
+      run: |
+        pytest tests/test_no_extra_install.py
+        coverage run -a -m pytest tests/test_no_extra_install.py
```
interesaaat, line 41: should we split this task in 2? Like run tests without coverage first, and coverage afterwards?

CONTRIBUTING.md
interesaaat: Good point!

ksaur:
Ready for review @interesaaat

Dockerfile
interesaaat: No I mean, is gools a typo?

README.md
```diff
@@ -15,10 +15,17 @@ Currently, you can use *Hummingbird* to convert your trained traditional ML mode
...
 Hummingbird was tested on Python >= 3.5 on Linux, Windows and MacOS machines.  It is recommended to use a virtual environment (See: [python3 venv doc](https://docs.python.org/3/tutorial/venv.html) or [Using Python environments in VS Code](https://code.visualstudio.com/docs/python/environments).)
+
+
+Install the Hummingbird package:
 ```
-pip install hummingbird-ml
+pip install hummingbird-ml[extra]
 ```
 
+This includes optional dependencies lightgbm and xgboost. If you do not need those, you can use:
```
interesaaat: Should we do the opposite? First having `pip install hummingbird-ml` (I think this should be the default) and then say, if you need all the extra dependencies such as Sklearn, LightGBM, etc, you can use: `pip install hummingbird-ml[extra]`.

hummingbird/ml/convert.py
```diff
@@ -54,7 +54,7 @@ def _convert_sklearn(model, test_input=None, extra_config={}):
         A model implemented in *PyTorch*, which is equivalent to the input *scikit-learn* model
     """
     assert model is not None
-    assert torch_installed(), "To use Hummingbird you need to install torch."
+    assert torch_installed(), "To use Hummingbird you need to install torch (or `pip install hummingbird-ml[extra]`)."
```
interesaaat: I think torch is not in extra.

tests/test_no_extra_install.py
```diff
@@ -0,0 +1,36 @@
+"""
+Tests Sklearn GradientBoostingClassifier converters.
```
interesaaat: This is a typo. Maybe something like `Test Hummingbird when no extra dependencies are installed`.

tests/test_no_extra_install.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+import unittest
+import warnings
+
+import numpy as np
+
+from hummingbird.ml._utils import sklearn_installed, lightgbm_installed, xgboost_installed
```
interesaaat: `sklearn_installed` is unnecessary here I think.

tests/test_no_extra_install.py
```diff
@@ -0,0 +1,36 @@
...
+    def test_lightgbm_installed_false(self):
+        warnings.filterwarnings("ignore")
+        assert not lightgbm_installed()
+
+    # Test no XGB returns false on xgboost_installed()
+    def test_xgboost_installed_false(self):
+        warnings.filterwarnings("ignore")
+        assert not xgboost_installed()
+
+    def test_import_convert_no_extra(self):
```
interesaaat, line 29: Can you please add a comment on this? Basically you should be able to still do import even if you don't have any extra, right?

Dockerfile
ksaur: LOL!  right.  i'll fix.  :-D

README.md
ksaur: Now that we have SKL as a req dep, yes, good point!

hummingbird/ml/convert.py
ksaur: thanks!

tests/test_no_extra_install.py
ksaur: good catch

tests/test_no_extra_install.py
ksaur: done

PR #123, opened 2020-06-11
Fix the documentation

interesaaat:


ahmedkrmn:
I was just about to open a new PR to add the hist-GBDT to the documentation.
Thanks for that!

PR #128, opened 2020-06-12
trying to get windows pipeline working

ksaur:


ksaur:
Wow https://github.com/pytorch/pytorch/issues/32186 is really a tough one

setup.py
```diff
@@ -26,13 +26,13 @@
 dependecy_links = []
 if sys.platform == "win32":
     if sys.version_info[:2] == (3, 5):
-        DEPENDENCY_LINKS = ["https://download.pytorch.org/whl/cpu/torch-1.5.0%2Bcpu-cp35-cp35m-win_amd64.whl"]
+        dependecy_links = ["https://download.pytorch.org/whl/cpu/torch-1.5.0%2Bcpu-cp35-cp35m-win_amd64.whl"]
```
ksaur: https://stackoverflow.com/questions/12518499/pip-ignores-dependency-links-in-setup-py

ksaur:
Ok, I think this is mostly ready to go after some struggles. 🤪   (this is going to def be a squash merge...🙈 )

For the LGBM install on MacOs, I put a note in troubleshooting; I don't want to mess with brew install within setup.py; at least i cant' find a good example.

@interesaaat is there anything else you want to bring over from the other branch?

setup.py
```diff
@@ -21,6 +22,29 @@
     if start_pos >= 0:
         long_description = long_description[start_pos:]
 
+install_requires = ["numpy>=1.15", "onnxconverter-common>=1.6.0", "scikit-learn==0.22.1"]
+if sys.platform == "darwin" or sys.platform == "linux":
+    install_requires.append("torch>=1.4.0")
+else:
+    if sys.version_info >= (3, 5, 0):
```
interesaaat: can we use the version from the github link you posted? not with dependency links.

ksaur: This try? 
https://github.com/microsoft/hummingbird/pull/128/commits/32cfc3b3be4fab4e5302acac0a1ddf6446ab1a5d
?

interesaaat: I think should be this `sys.version_info[:2] == (3, 5):`

ksaur: trying.....

PR #86, opened 2020-05-30
Add ONNX as input and output model format

interesaaat:
The support for ONNX is just preliminary and for the moment tested only on ONNXML models generated by the [LightGBM converter](https://github.com/onnx/onnxmltools/tree/master/onnxmltools/convert/lightgbm) in ONNXMLTOOLS.

I also did some refactoring to make the process of adding backends more scalable and constrained.

PR #142, opened 2020-06-16
Add ONNX support

interesaaat:
This PR tries to fix the coverage problem of #86.


.github/workflows/pythonapp.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         pip install .[tests]
     - name: Run basic tests without extra
       run: |
-        pytest tests/test_no_extra_install.py
+        pytest
```
ksaur, line 36: uhhh how is this working?  These should fail because the rest of the libraries aren't installed

.github/workflows/pythonapp.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         export LDFLAGS="$LDFLAGS -Wl,-rpath,/usr/local/opt/libomp/lib -L/usr/local/opt/libomp/lib -lomp"
     - name: Install extra dependencies
       run: |
-        pip install .[extra]
+        pip install  --no-cache-dir .[extra,onnx]
```
ksaur: i think i determined that we don't need `--no-cache-dir` here. was it giving some error?

tests/test_onnxml_converter_lightgbm.py
```diff
@@ -0,0 +1,302 @@
...
+        np.testing.assert_allclose(
+            list(map(lambda x: list(x.values()), onnx_ml_pred[0])), onnx_pred[0], rtol=rtol, atol=atol
+        )  # probs
+
+    # Check that ONNXML models can only target the ONNX backend.
+    @unittest.skipIf(
+        not (onnx_ml_tools_installed() and onnx_installed()), reason="ONNXML test require ONNX, ORT and ONNXMLTOOLS"
+    )
+    def test_lightgbm_pytorch(self):
+        X = [[0, 1], [1, 1], [2, 0]]
```
ksaur, line 83: Does datasize impact anything we need to test?

hummingbird/ml/convert.py
```diff
@@ -27,14 +27,41 @@
 from .supported import lgbm_operator_list  # noqa
 
 
+def _is_onnx_model(model):
+    """
+    Function returning whether the input model is an ONNX model or not.
+    """
+    return "onnx" in model.__module__ and "graph" in dir(model)
```
ksaur: maybe something like

```
>>> type(onnx_ml_model).__name__
'ModelProto'
```
or
```
>>> isinstance(onnx_ml_model, onnx.onnx_ONNX_REL_1_6_ml_pb2.ModelProto)
True
```

ksaur: i guess for the second one, the version isn't known. but we could check just the first part

.github/workflows/pythonapp.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         pytest
     - name: Coverage
       run: |
-        coverage run -a -m pytest tests
+        coverage run -a -m pytest tests --ignore=tests/test_no_extra_install.py
```
ksaur: and this part should no longer be needed due to the skips, although it doesn't hurt.  (maybe you just did a revert on this file?)

hummingbird/ml/convert.py
```diff
@@ -56,13 +83,15 @@ def _convert_sklearn(model, test_input=None, extra_config={}):
...
+
     # Parse scikit-learn model as our internal data structure (i.e., Topology)
-    # We modify the scikit learn model during optimizations.
+    # We modify the scikit learn model during translation.
     model = deepcopy(model)
     topology = parse_sklearn_api_model(model)
 
     # Convert the Topology object into a PyTorch model.
-    hb_model = _convert_topology(topology, extra_config=extra_config)
+    hb_model = topology_converter(topology, extra_config=extra_config)
```
ksaur, line 94: question: this is now public because the control flow changed?

hummingbird/ml/ir_converters/linked_node.py
```diff
@@ -0,0 +1,314 @@
...
+from onnx import helper, shape_inference
+import torch
+
+from ..operator_converters import constants
+from ..supported import get_onnxml_api_operator_name
+from ..exceptions import MissingConverter
+
+
+def convert(
+    node_list, input_tensors, initializers, output_names, test_data, output_model_name=None, target_opset=9, extra_config={},
```
ksaur, line 28: Why 9 and not 11?  (is there some particular operator we use?)

ksaur (commented):
initial comments, more tomorrow ;)

.github/workflows/pythonapp.yml
interesaaat: I added conditions on all tests. I think it is better since we cannot expect users to know which test to run or not run based on their dependencies.

hummingbird/ml/convert.py
```diff
@@ -135,73 +164,147 @@ def _convert_xgboost(model, test_input, extra_config={}):
...
-        except ValueError:
-            raise MissingConverter(
-                "Unable to find converter for {} type {} with extra config: {}.".format(
-                    operator.type, type(getattr(operator, "raw_model", None)), extra_config
+    # We modify the model during translation.
+    model = deepcopy(model)
+
+    # Parse an ONNX-ML model into our internal data structure (i.e., LinkedNode)
+    graph = model.graph
+    input_names = input_names if input_names is not None else [in_.name for in_ in graph.input]
```
ksaur, line 220: I'm a little confused about what the difference is between `input_names` and `in`?

hummingbird/ml/convert.py
```diff
@@ -135,73 +164,147 @@ def _convert_xgboost(model, test_input, extra_config={}):
...
+    For *Sklearn*, *LightGBM* and *XGBoost* currently only the *torch* backend is supported.
+    For *ONNX* currently only the *onnx* backend is supported. For ONNX models, Hummingbird behave as a model
+    rewriter converting [ONNX-ML] operators into [ONNX ones].
     The detailed list of models and backends can be found at `hummingbird.ml.supported`.
     [Sklearn]: https://scikit-learn.org/
     [LightGBM]: https://lightgbm.readthedocs.io/
     [XGBoost]: https://xgboost.readthedocs.io/
+    [ONNX]: https://onnx.ai/
+    [ONNX-ML]: https://github.com/onnx/onnx/blob/master/docs/Operators-ml.md
+    [ONNX ones]: https://github.com/onnx/onnx/blob/master/docs/Operators.md
```
ksaur: "ones" is confusing here

hummingbird/ml/convert.py
```diff
@@ -135,73 +164,147 @@ def _convert_xgboost(model, test_input, extra_config={}):
...
         backend: The target for the conversion
-        test_input: some input data used to trace the model execution
+        test_input: some input data used to trace the model execution.
+                    For the ONNX backend the test_input size is supposed to be as large as the expected batch size.
         extra_config: Extra configurations to be used by the individual operator converters.
                       The set of supported extra configurations can be found at `hummingbird.ml.supported`
 
     Examples:
-        >>> pytorch_model = convert(sklearn_model,`pytorch`)
+        >>> pytorch_model = convert(sklearn_model,`torch`)
```
ksaur, line 290: We still need some comment somewhere (not necessarily here) about how these are equivalent, and that this is a non-breaking change.  I'm being picky because folks have >100 forks of HB, and I don't want to break anything or confuse anyone like last time w/pypi! (our blog, etc has this as 'pytorch' so i just want to make sure we communicate this change and that it is backward compat.)

hummingbird/ml/ir_converters/linked_node.py
```diff
@@ -0,0 +1,314 @@
...
+        output_model_name = str(uuid4().hex) + ".onnx"
+    onnx_model_name = output_model_name
+    # Container for the output IR operators. We will be using this to generate the final ONNX model.
+    output_onnx_ir = []
+    # During translation we need to make sure that each variable is unique across operators.
+    variable_check = set()
+    initializer_tensors = {}
+    output_tensors = {}
+
+    assert len(input_tensors) == 1, "Hummingbird ONNX converter currently support only graphs with a single input."
```
ksaur: "currently support_**s**_ only graphs"

hummingbird/ml/ir_converters/linked_node.py
```diff
@@ -0,0 +1,314 @@
...
+                if len(test_output_names) == 1 and next(iter(test_output_names)) == input_tensors[0].name:
+                    # This is the first operator: use the input parameter.
+                    inputs = torch.from_numpy(test_data)
+                else:
+                    # We are in the middle of a graph: generate inputs using the part of the graph converted so far.
+                    # Need to be tested once we get more complex ONNX models.
+                    raise RuntimeError(
+                        "Inference of ONNX models with multiple ONNXML operators is not supported yet. Please fill an issue at https://github.com/microsoft/hummingbird."
+                    )
+                    # graph = helper.make_graph(
```
ksaur, line 107: Maybe add a comment here that these lines are commented out because they are future work

hummingbird/ml/operator_converters/onnx_tree_ensembles.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def convert_onnx_tree_enseble_regressor(operator, device=None, extra_config={}):
+    """
+    Converter for `onnxml.TreeEnsembleRegressor`.
+
+    Args:
+        operator: An operator wrapping a `onnxml.TreeEnsembleRegressor` model
+        device: String defining the type of device the converted operator should be run on
+        extra_config: Extra configuration used to select the best conversion strategy
```
ksaur, line 175: Do you document somewhere what the possibilities for `extra_config` are?

.github/workflows/pythonapp.yml
ksaur: Hmmm...but now the tests look really cluttered, but I guess it's just the onnx/xbg/lgbm tests that look cluttered.  Maybe we should consider some sort of different directory structure going forward.

.github/workflows/pythonapp.yml
interesaaat: Nope, I just copied from the version we had before. Let me remove it.

tests/test_onnxml_converter_lightgbm.py
interesaaat: I think I have a mix of test using random numbers and fixed arrays for that reason. The fixed array tests are taken from the LightGBM convert tests.

tests/test_backends.py
```diff
@@ -25,7 +30,7 @@ def test_backends_case_sensitive(self):
 
         model.fit(X, y)
 
-        hb_model = hummingbird.ml.convert(model, "pYtOrCh")
+        hb_model = hummingbird.ml.convert(model, "tOrCh")
```
ksaur, line 33: Can we also add a test that "pytorch" still works?

hummingbird/ml/convert.py
interesaaat: Yes the version can change. Should I go with the first one?

.github/workflows/pythonapp.yml
interesaaat: yep

interesaaat: My doubt was whether coverage will get messed up since we generate the coverage when some dependencies are off and then again when they are on, and we run the same tests.

hummingbird/ml/operator_converters/_gbdt_commons.py
```diff
@@ -41,6 +41,10 @@ def convert_gbdt_classifier_common(tree_infos, get_tree_parameters, n_features,
         n_classes -= 1
     if classes is None:
         classes = [i for i in range(n_classes)]
+    reorder_trees = True
+    if constants.REORDER_TREES in extra_config:
```
ksaur, line 45: I'm confused about why this is in this PR.  Does it relate to onnx, or part of some other refactoring?

hummingbird/ml/convert.py
interesaaat: Yes before it was a private method inside convert.py, but now it's a method in its own file. I wanted to call the convert method for each IR the same. Maybe I can have all the IRs subclassing an abstract class with a convert method. I can try to have them using the signature using extra_config or `**kwargs`

hummingbird/ml/operator_converters/onnx_tree_ensembles.py
```diff
@@ -0,0 +1,186 @@
...
+"""
+Converters for ONNX-ML tree-ensemble models.
+"""
+
+import numpy as np
+from onnxconverter_common.registration import register_converter
+
+from . import constants
+from ._gbdt_commons import convert_gbdt_classifier_common, convert_gbdt_common
+from ._tree_commons import TreeParameters, get_parameters_for_tree_trav_common, convert_decision_ensemble_tree_common
```
ksaur: I don't see where either of these is used in this file, unless I am missing it.  But it passed lint/flake8...so maybe I don't understand something.

hummingbird/ml/ir_converters/linked_node.py
interesaaat: 9 is default in the torch to onnx converter.

hummingbird/ml/operator_converters/onnx_tree_ensembles.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    Function used to extract the parameters from a ONNXML TreeEnsemble model.
+    """
+    tree_infos = []
+    left = right = features = values = threshold = None
+    tree_ids = target_node_ids = target_tree_ids = modes = None
+    classes = post_transform = None
+
+    for attr in model.origin.attribute:
+        if attr.name == "nodes_falsenodeids":
```
ksaur, line 33: Maybe put a comment about how these strings are from https://xadupre.github.io/skl2onnx/onnx_ops.html or something

hummingbird/ml/operator_converters/skl_decision_tree.py
```diff
@@ -47,31 +40,10 @@ def convert_sklearn_random_forest_classifier(operator, device, extra_config):
     if not all(isinstance(c, int) for c in classes):
         raise RuntimeError("Random Forest Classifier translation only supports integer class labels")
 
-    tree_parameters, max_depth, tree_type = get_tree_params_and_type(
-        tree_infos, get_parameters_for_sklearn_common, extra_config
+    return convert_decision_ensemble_tree_common(
+        tree_infos, get_parameters_for_sklearn_common, get_parameters_for_tree_trav_sklearn, n_features, classes, extra_config
     )
 
-    # Generate the tree implementation based on the selected strategy.
```
ksaur, line 54: were these moved to make sense with the onnxml trees as well? I am a bit lost here and need more info.

hummingbird/ml/supported.py
```diff
@@ -93,11 +95,38 @@ def _build_lightgbm_operator_list():
...
     The set of supported backends is defined here.
     """
-    return {"pytorch"}
+    backends = set()
+
+    if torch_installed():
+        import torch
+
+        backends.add(torch.__name__)
+        backends.add("py" + torch.__name__)  # For compatibility with earlier versions.
```
ksaur, line 156: thanks! ;)

ksaur (commented):
next batch....

hummingbird/ml/convert.py
ksaur: yah, i think the first one is cleanest, although i don't feel strongly :)

hummingbird/ml/convert.py
ksaur: i think it's ok as-is, i just wanted to make sure i understood why the change. (and maybe something fancier later on)

.github/workflows/pythonapp.yml
interesaaat: Yes, agreed.

hummingbird/ml/convert.py
interesaaat: Do you think that `input` is better than `in_`? Although maybe create confusions with `inputs`

hummingbird/ml/convert.py
interesaaat: Suggestions? ONNX is already taken :)

interesaaat: Ok I changed it with [ONNX operators]

hummingbird/ml/convert.py
interesaaat: We can add an item on this in the next release description.

hummingbird/ml/ir_converters/linked_node.py
interesaaat: In addition to line 102?

hummingbird/ml/operator_converters/onnx_tree_ensembles.py
interesaaat: In convert we have `extra_config: Extra configurations to be used by the individual operator converters.
                      The set of supported extra configurations can be found at hummingbird.ml.supported`. Do you think it's ok? This `extra_config` however can be a bit different since they are into the converter and only for developers.

tests/test_backends.py
interesaaat: yep

hummingbird/ml/operator_converters/_gbdt_commons.py
interesaaat: yes this is something needed for onnx and so that we can hook into the gbdt code.

hummingbird/ml/operator_converters/onnx_tree_ensembles.py
interesaaat: Yea I don't know what is going on with flake8. I removed the unnecessary one.

hummingbird/ml/operator_converters/onnx_tree_ensembles.py
interesaaat: Good point! Let me add a link.

hummingbird/ml/operator_converters/skl_decision_tree.py
interesaaat: We already merged a PR with this changes actually (#132).

hummingbird/ml/convert.py
ksaur: ooh i see what you're doing here now.  it was just confusing at first.

ksaur (approved):
LGTM, waiting to merge #145 first.

interesaaat:
If the tests pass it should be good to go. They just bumped PyTorch to 1.5.1. There is a bug in the ONNX exporter for GEMM: I thought they fixed it by with 1.5.1 it's still failing. I enabled `tree_trav` by default.

PR #145, opened 2020-06-17
Cleanup Normalizer Converter Code

### LINKED ISSUE #126: Cleanup Normalizer Converter Code
We have the original draft of the normalizer code uploaded in [this branch](https://github.com/microsoft/hummingbird/tree/cleanup/normalizer). All credit here goes to @scnakandala, the original author and brains behind this.

It contains an un-edited implementation of SklearnNormalizer [here](https://github.com/microsoft/hummingbird/blob/cleanup/normalizer/hummingbird/ml/operator_converters/normalizer.py).

There is a single test file [here](https://github.com/microsoft/hummingbird/blob/cleanup/normalizer/tests/test_sklearn_normalizer_converter.py) that needs to be cleaned up and passing.

ahmedkrmn:
Closes #126 

This PR cleans up the implementation of the [sklearn.preprocessing.Normalizer](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.Normalizer.html) as well as its test file from the [cleanup/normalizer](https://github.com/microsoft/hummingbird/tree/cleanup/normalizer) branch.

### DIFF
--- a/hummingbird/ml/operator_converters/normalizer.py
+++ b/hummingbird/ml/operator_converters/normalizer.py
@@ -1,0 +1,31 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import torch
+from onnxconverter_common.registration import register_converter
+
+
+class Normalizer(torch.nn.Module):
+    def __init__(self, norm, device):
+        super(Normalizer, self).__init__()
+        self.norm = norm
+        self.regression = False
+
+    def forward(self, x):
+        if self.norm == "l1":
+            return x / torch.abs(x).sum(1, keepdim=True)
+        elif self.norm == "l2":
+            return x / torch.pow(torch.pow(x, 2).sum(1, keepdim=True), 0.5)
+        elif self.norm == "max":
+            return x / torch.max(torch.abs(x), dim=1, keepdim=True)[0]
+        else:
+            raise RuntimeError("Unsupported norm: {0}".format(self.norm))
+
+
+def convert_sklearn_normalizer(operator, device, extra_config):
+    return Normalizer(operator.raw_operator.norm, device)
+
+
+register_converter("SklearnNormalizer", convert_sklearn_normalizer)
--- a/doc/html/hummingbird/ml/supported.html
+++ b/doc/html/hummingbird/ml/supported.html
@@ -27,6 +27,7 @@
 <p><strong>Supported Backends</strong>
 PyTorch</p>
 <p><strong>Supported Operators</strong>
+Normalizer,
 DecisionTreeClassifier,
 DecisionTreeRegressor,
 ExtraTreesClassifier,
@@ -58,6 +59,7 @@ All operators, backends, and configurations settings supported in Hummingbird ar
 PyTorch
 
 **Supported Operators**
+Normalizer,
 DecisionTreeClassifier,
 DecisionTreeRegressor,
 ExtraTreesClassifier,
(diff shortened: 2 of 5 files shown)


ahmedkrmn:
Hi @ksaur,
Do you think I should add relative and absolute tolerance in the test?
Also is there a way to automatically generate an HTML file for the normalizer converter similar to [gbdt.html](https://github.com/microsoft/hummingbird/blob/master/doc/html/hummingbird/ml/operator_converters/gbdt.html) or is it written manually?

ksaur:
Thanks for your PR!

No worries on the HTML doc for now; we are working on automating that process and it will be generated at the next release


For`tests/test_sklearn_normalizer_converter.py ` line 25  you should have:

```
np.testing.assert_allclose(model.transform(data), pytorch_model(torch.from_numpy(data)).data.numpy(), rtol=1e-06, atol=1e-06)
```
 
But…when I tried this with your code (to make sure I was telling you correctly), I saw that I got an infinite loop/`StopIteration` inside `_container.py`.  This led to us doing a quick bugfix just now.   (We hadn't hit this bug before because normalizer has no param, while the other models have the param)

So hopefully if you merge the changes from PR #146, it should "just work".

hummingbird/ml/operator_converters/normalizer.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def forward(self, x):
+        if self.norm == "l1":
+            return x / torch.abs(x).sum(1, keepdim=True)
+        elif self.norm == "l2":
+            return x / torch.pow(torch.pow(x, 2).sum(1, keepdim=True), 0.5)
+        elif self.norm == "max":
+            return x / torch.max(torch.abs(x), dim=1, keepdim=True)[0]
+        else:
+            raise RuntimeError("Unsupported norm: {0}".format(self.norm))
```
ksaur, line 24: Please also write a test for `raise RuntimeError("Unsupported norm: {0}".format(self.norm))`.
you can see some example of error test cases [here](https://github.com/microsoft/hummingbird/blob/master/tests/test_sklearn_gbdt_converters.py#L149).

ksaur (changes requested):

ahmedkrmn:
> np.testing.assert_allclose(model.transform(data), pytorch_model(torch.from_numpy(data)).data.numpy(), rtol=1e-06, atol=1e-06)
> ```
>  
Do you think it's necessary to convert the tensor data to a numpy array? `assert_allclose` works with array_like data structures so they work with tensors out of the box.

interesaaat:
> > np.testing.assert_allclose(model.transform(data), pytorch_model(torch.from_numpy(data)).data.numpy(), rtol=1e-06, atol=1e-06)
> > ```
> >  
> > ```
> 
> Do you think it's necessary to convert the tensor data to a numpy array? `assert_allclose` works with array_like data structures so they work with tensors out of the box.

Yes no need to do it. On the result instead you should do `.cpu().numpy()`.

ksaur:
Sorry I pasted the old code; I meant
```
np.testing.assert_allclose(model.transform(data), pytorch_model(data).cpu().numpy(), rtol=1e-06, atol=1e-06)

```

ahmedkrmn:
Hi Karla,
I've added a test for RuntimeError as requested. I've also changed the `data` array in `test_normalizer_converter` to be randomly generated.
I've question, I've noticed that the other tests have this line in functions that raise errors:
```python
warnings.filterwarnings("ignore")
```
What is this used for, and do we need to add it in out case?

ksaur:
`warnings.filterwarnings("ignore")` was for when we were getting some warnings from `scikit-learn`, but I believe that was when we were working with `0.21.3`.  (We've since moved to SKL `0.22.1` and I *think* the warnings went away but I'd have to double check.)  If you're not seeing third-party warnings, then it's fine.

ksaur (approved):
LGTM

PR #149, opened 2020-06-18
Update LGBM-example.ipynb

jspisak:
Update to add !pip install at the beginning of the NB. This will reduce the tracebacks seen by colab users.


ksaur:
In the next release we need to have the [extra] tag for lgbm/xgb on the install since they won't be installed as dependency by default. will merge right before pushing changes to pypi

PR #173, opened 2020-06-27
add LGBMRanker

### LINKED ISSUE #172: Support for LGBMRanker?
Hello there, 

I am wondering if the current implementation supports convertion from LightGBMRanker which is part of the LightGBM package? Is there anything special to the LightGBMRanker and will the converter for LGBMRegressor works in this case? Here is the error I got when I try to convert the LGBMRanker

```
MissingConverter: Unable to find converter for model type <class 'lightgbm.sklearn.LGBMRanker'>.
It usually means the pipeline being converted contains a
transformer or a predictor with no corresponding converter implemented.
Please fill an issue at https://github.com/microsoft/hummingbird.
```

This is the doc to LGBMRanker https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMRanker.html

Thanks in advance!

ksaur:
Adding `LGBMRanker` for Issue #172

### DIFF
--- a/tests/test_lightgbm_converter.py
+++ b/tests/test_lightgbm_converter.py
@@ -16,7 +16,7 @@ if lightgbm_installed():
 
 class TestLGBMConverter(unittest.TestCase):
     # Check tree implementation
-    @unittest.skipIf(not lightgbm_installed(), reason="LightGBM test require LightGBM installed")
+    @unittest.skipIf(not lightgbm_installed(), reason="LightGBM test requires LightGBM installed")
     def test_lgbm_implementation(self):
         warnings.filterwarnings("ignore")
         np.random.seed(0)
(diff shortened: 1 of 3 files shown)


interesaaat:
This was quick!

tests/test_lightgbm_converter.py
```diff
@@ -50,45 +50,81 @@ def _run_lgbm_classifier_converter(self, num_classes, extra_config={}):
...
+            model.fit(X, y, group=[X.shape[0]], eval_set=[(X, y)], eval_group=[X.shape[0]])
+
+            torch_model = hummingbird.ml.convert(model, "torch", extra_config=extra_config)
+            self.assertIsNotNone(torch_model)
+            np.testing.assert_allclose(model.predict(X), torch_model.predict(X), rtol=1e-06, atol=1e-06)
+
+    # Ranker
+    @unittest.skipIf(not lightgbm_installed(), reason="LightGBM test requires LightGBM installed")
+    def test_lgbm_binary_ranker_converter(self):
+        # Using 30 rather than 1000 to avoid "label (31) excel the max range" error
```
interesaaat: What is this? Any idea?

ksaur: I followed the instructions to set `label_gain` without luck https://github.com/microsoft/LightGBM/issues/1090

ksaur: I can try again Monday! Maybe there is something minor I was missing

interesaaat: Ok let's leave this test as is. Can you add one more with label_gain once you figure it out? I will then merge it!

ksaur: figured out `label_gain=list(range(1000))`. should be ok now i think

PR #175, opened 2020-06-29
[WIP] Adding onnxml linear converters

### LINKED ISSUE #157: add converters for onnx.ml->pyt for linear models
Now that we have skl->pyt done, we need to have corresponding onnx.ml -> pyt converters

ksaur:
Waiting on: Inference of ONNX models with multiple ONNXML operators is not supported yet.


interesaaat:
For the moment please select ONNX as output. What's the problem with the registration? It should be as sklearn models.

PR #185, opened 2020-07-09
Adding long description to setup.py

### LINKED ISSUE #184: Add long_description to setup.py
For some reason the l`ong_description` field is generate but not set in `setup.py`. This will also give a nicer presentation on pypi.

interesaaat:
Closes #184.


ksaur:
Awesome,  i noticed  that as well, thanks for this!

Issue #174, opened 2020-06-29
Support for XGBRanker


ksaur:
[XGBRanker](https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBRanker) should be a very simple add; similar to #173 

It should hopefully just involve (1) either using or slightly tweaking the converter for XGBRegressor (2) writing tests and making sure they pass

KranthiGV:
I was looking at this a couple of days ago. Is it okay if I pick this up?

interesaaat:
Go for it!

ksaur:
closed with #189

PR #189, opened 2020-07-10
Add XGBRanker as a supported operator

KranthiGV:
This addresses #174 

In this PR:
- Added XGBoost ranker (Sklearn API) as a supported operator.
- Also, added a couple of tests around the same.

### DIFF
--- a/hummingbird/ml/operator_converters/xgb.py
+++ b/hummingbird/ml/operator_converters/xgb.py
@@ -103,23 +103,6 @@ def convert_sklearn_xgb_classifier(operator, device, extra_config):
     return convert_gbdt_classifier_common(tree_infos, _get_tree_parameters, n_features, n_classes, extra_config=extra_config)
 
 
-def convert_sklearn_xgb_ranker(operator, device, extra_config):
-    """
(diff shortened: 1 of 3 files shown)


hummingbird/ml/operator_converters/xgb.py
```diff
@@ -103,6 +103,23 @@ def convert_sklearn_xgb_classifier(operator, device, extra_config):
     return convert_gbdt_classifier_common(tree_infos, _get_tree_parameters, n_features, n_classes, extra_config=extra_config)
 
 
+def convert_sklearn_xgb_ranker(operator, device, extra_config):
```
interesaaat: I think that we can just call `convert_sklearn_xgb_regressor` from `register_converter`.

interesaaat (commented):
Thanks for fixing the typos!

hummingbird/ml/operator_converters/xgb.py
KranthiGV: I thought it'd be easier to read this way. Anyways, let me change that.

KranthiGV:
I think it looks good now. Please have a look!

interesaaat (approved):

Issue #178, opened 2020-07-06
Support for Isolation Forest


zhanjiezhu:
Hi there, thanks for the great package, I would like to work on the support for Isolation Forest sklearn implementation (https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.IsolationForest.html). Since this implementation has 2 main methods "predict" and "decision_function", where the "decision_function" returns the anomaly scores, and "predict" returns either -1 (anomaly) or 1 (normal) based on comparing the anomaly score with an "offset_".

I would propose following changes in order to support Isolation Forest:
1. Replace "value" in each tree with (normalized) depth, since the existing value in each tree is just useless random numbers due to the randomness when growing each tree in Isolation Forest. To do this, I will create a separate "convert_sklearn_isolation_forest" and "_get_parameters_for_sklearn_iforest" to extract and change the tree parameters. I have already done this step and it's returning expected results. 
2. create a new "PyTorchBackendModelAnomalyDetection", in which I will implement "predict" and "decision_function", and I will also need to use the "offset_" in extra_config to normalize the final output. To be able to call "PyTorchBackendModelAnomalyDetection", I will also need to add a boolean flag "anomaly_detection" just like "regression".

I've already done the first step, but would like to get your feedback before continuing.

Thanks!

interesaaat:
Hey @zhanjiezhu this looks great! I assigned the issue to you.

PR #191, opened 2020-07-12
Support isolation forest in sklearn

zhanjiezhu:
Addresses issue #178 

This PR:
1. Added converter for sklearn isolation forest, including tree implementation for isolation forest (e.g. GEMMIsolationForestImpl), which uses constants e.g. offset to normalize the anomaly scores
2. Added "PyTorchBackendModelAnomalyDetection", which supports 3 methods of sklearn anomaly detection, i.e. predict, decision_function & score_samples
3. Added a new param "is_anomaly_detection" to the base tree implementation classes (e.g. AbstractPyTorchTreeImpl) in order to differentiate from regression & classification mode.
4. Added tests.


hummingbird/ml/_container.py
```diff
@@ -99,3 +100,26 @@ def predict_proba(self, *inputs):
         On classification tasks returns the probability estimates.
         """
         return self.forward(*inputs)[1].cpu().numpy()
+
+
+class PyTorchBackendModelAnomalyDetection(PyTorchBackendModel):
```
interesaaat: What about extending `PyTorchBackendModelRegression` rather then `PyTorchBackendModel`? You can then add the comment `On anomaly detection (e.g. isolation forest) returns the predicted classes (-1 or 1).` directly ofter `On classification...`

hummingbird/ml/operator_converters/_tree_implementations.py
```diff
@@ -67,13 +67,14 @@ class AbstractPyTorchTreeImpl(AbstracTreeImpl, torch.nn.Module):
     Abstract class definig the basic structure for tree-base models implemented in PyTorch.
     """
 
-    def __init__(self, tree_parameters, n_features, classes, n_classes):
+    def __init__(self, tree_parameters, n_features, classes, n_classes, is_anomaly_detection):
```
interesaaat: Rather that do this, can you please add `anomaly_detection` to `BaseOperator`? We use that class for this purpose.

hummingbird/ml/ir_converters/topology.py
```diff
@@ -46,6 +47,10 @@ def convert(topology, device=None, extra_config={}):
     if operator_map[operators[-1].full_name].regression:
         # We are doing a regression task.
         pytorch_container = PyTorchBackendModelRegression
+    elif hasattr(operator_map[operators[-1].full_name], 'is_anomaly_detection') and \
```
interesaaat: If you add `anomaly_detection` to `BaseOperator` (look at the comment below) you can remove the `hasattr`.

hummingbird/ml/operator_converters/skl_iforest.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def _average_path_length(n_samples_leaf):
+    """
+    Taken from sklearn implementation of isolation forest:
+    https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/ensemble/_iforest.py#L480
+    The average path length in a n_samples iTree, which is equal to
+    the average path length of an unsuccessful BST search since the
+    latter has the same structure as an isolation tree.
+    Parameters
```
interesaaat: Can you refactor this doc a bit to follow the pdoc convention: Like using `Args:` and `Return:`

interesaaat (changes requested):
Great job! I mostly have stylistic comments.

hummingbird/ml/_container.py
zhanjiezhu: make sense, will do!

hummingbird/ml/operator_converters/_tree_implementations.py
zhanjiezhu: Make sense! I totally missed the BaseOperator, indeed I can add that. But I wonder if I can remove those "anomaly_detection" arguments in those base implementations, because in `__init__` of `AbstractPyTorchTreeImpl` there is a condition on self.anomaly_detection, so if I remove all those arguments, it will always be False because it calls the "__init__" of BaseOperator first. What do you think?

hummingbird/ml/ir_converters/topology.py
zhanjiezhu: will do!

hummingbird/ml/operator_converters/skl_iforest.py
zhanjiezhu: will do

hummingbird/ml/operator_converters/_tree_implementations.py
interesaaat: Yea I was checking the code and I don't think we can completely remove it, you are right. I was thinking that maybe we can move these code directly in the implementations in the `skl-iforest` file but I am not sure it will actually be better.

zhanjiezhu: I can either keep all those argument, or maybe do this: remove all those arguments, add a line `self.anomaly_detection = True` in each isolation forest implementation right before the `super().__init__()`, then in the BaseOperator, add a condition: `if not hasattr(self, 'anomaly_detection'): self.anomaly_detection = False`.

Which one do you think is better?

interesaaat: Good idea! Do we actually need `if not hasattr(self, 'anomaly_detection'): self.anomaly_detection = False`? Since `BaseOperator` is called first, we can just have `anomaly_detection = False` by default and simply set it in the subclass, no?

zhanjiezhu: I don't think so, that's actually what I did at first then found out not working, there could be 2 situations: 1. if I set it True in the subclass **before** the `super().__init__()`, then it will be overwritten by `BaseOperator` and always False; 2. if I set it True in the subclass **after** the `super().__init__()`, then it can't change the condition in e.g. `AbstractPyTorchTreeImpl` because the `BaseOperator` is called first and had anomaly_detection = False. So I have to choose the 1st situation, and also need to add a condition to avoid being overwritten by `BaseOperator` if it's already assigned value. What do you think?

interesaaat: Ok I see your point now. What if we change the signature of `BaseOperator` and have it taking something like `regression=False, classification=False,anomaly_detection=False`. In this case we can directly set the behavior from the base class is known in advance before `__init__` or set if directly afterward.

zhanjiezhu: Indeed good idea to add the default parameters in `BaseOperator`, but don't understand how the last sentence can work. Now the class inheritance is like: `BaseOperator` => `AbstracTreeImpl` => `AbstractPyTorchTreeImpl` => `GEMMTreeImpl` => `GEMMIsolationForestImpl`. In `AbstractPyTorchTreeImpl` I need to do condition based on `anomaly_detection` which is default to be False in `BaseOperator` and I want to set it True in the last subclass `GEMMIsolationForestImpl`. I don't see how I can do that by simply adding the default parameters, maybe you mean something like `super().super().super()...__init__(anomaly_detection=True)`? But still it will be overwritten later when calling the `__init__`. And how about this: add default parameters in `BaseOperator` as you suggested, also add a `**kwargs` in parameters and __init__ calls in the base and every subclass in the chain, so that in the final subclass I can simply set that to True, and for other types of models it will be default to False if not specified. This should allow adding other parameters easier in the future as well, what do you think?

interesaaat: Deal! :) I think this is a good mix between ease of implementation and generality.

zhanjiezhu: great! will push new commits later

interesaaat:
I commit few style fixes. Otherwise looks good! I will wait for all the checks to conclude and I will merge afterwards. Thanks @zhanjiezhu!

interesaaat (approved):

PR #193, opened 2020-07-14
one hot encoder

### LINKED ISSUE #136: Cleanup and import SKL Feature Pre-processing: OneHotEncoder
We have the original draft of the OneHotEncoder code uploaded in this [branch](https://github.com/microsoft/hummingbird/tree/cleanup/one-hot-enc). All credit here goes to @scnakandala, the original author and brains behind this.

It contains an un-edited implementation of OneHotEncoder [here](https://github.com/microsoft/hummingbird/blob/cleanup/one-hot-enc/hummingbird/ml/operator_converters/one_hot_encoder.py).

There is a single test file [here](https://github.com/microsoft/hummingbird/blob/cleanup/one-hot-enc/tests/test_sklearn_one_hot_encoder_converter.py) that needs to be cleaned up and passing.

Notes:
* We had to treat strings a bit differently because of the way they are represented in tensors
  *  We should add more robust tests here
* the `retain_order`  flag in `OneHotEncoder` is not yet implemented. We can wait on that for later unless someone gets inspired.

ksaur:
scikit-learn one-hot-encoder impl
Scikit learn had API change that removed categorical_features, complicating things a bit


hummingbird/ml/operator_converters/skl_one_hot_encoder.py
```diff
@@ -0,0 +1,82 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import torch
+import numpy as np
+from onnxconverter_common.registration import register_converter
```
interesaaat, line 8: Alphabetic order here please!

hummingbird/ml/operator_converters/skl_one_hot_encoder.py
```diff
@@ -0,0 +1,82 @@
...
+                1, -1, self.max_word_length // 4
+            )
+            encoded_tensors.append(torch.prod(torch.eq(x[:, i : i + 1, :], conditions), dim=2))
+
+        return torch.cat(encoded_tensors, dim=1).float()
+
+
+class OneHotEncoder(BaseOperator, torch.nn.Module):
+    def __init__(self, categories, device):
+        super(OneHotEncoder, self).__init__()
```
interesaaat: I think that if you pass trasformer = True here you can avoid setting it explicitly in line 59. (We should actually fix this in the other operators too, we just changed the signature in the iforest PR).

hummingbird/ml/operator_converters/skl_one_hot_encoder.py
```diff
@@ -0,0 +1,82 @@
...
+        self.condition_tensors = torch.nn.ParameterList(condition_tensors)
+        self.transformer = True
+
+    def forward(self, x):
+        if x.dtype != torch.int64:
+            x = x.long()
+
+        return self.transform(x)
+
+    def transform(self, x):
```
interesaaat: I don't think this function is used elsewhere, can we move the code directly in forward?

hummingbird/ml/operator_converters/skl_one_hot_encoder.py
```diff
@@ -0,0 +1,82 @@
...
+                np.array(arr, dtype="|S" + str(self.max_word_length))
+                .view("int32")
+                .reshape(-1, self.max_word_length // 4)
+                .tolist()
+            )
+            condition_tensors.extend(cats)
+            num_categories.append(num_categories[-1] + len(cats))
+        self.condition_tensors = torch.nn.Parameter(torch.IntTensor(condition_tensors), requires_grad=False)
+        self.num_categories = num_categories
+        self.transformer = True
```
interesaaat: Same this here (see comment below).

hummingbird/ml/operator_converters/skl_one_hot_encoder.py
```diff
@@ -0,0 +1,82 @@
...
+# license information.
+# --------------------------------------------------------------------------
+import torch
+import numpy as np
+from onnxconverter_common.registration import register_converter
+
+from ._base_operator import BaseOperator
+
+
+class OneHotEncoderString(BaseOperator, torch.nn.Module):
```
interesaaat, line 14: Do you feel to add some documentation in this class?

ksaur: I added doc to the top of the function. Inside the function, feel free to add more!

interesaaat:
Ok I added few comments. Will merge once all checks pass

interesaaat:
Closes #136.

PR #196, opened 2020-07-15
onnx scaler

ksaur:


hummingbird/ml/operator_converters/_scaler_implementations.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class Scaler(BaseOperator, torch.nn.Module):
+    """
+    Class implementing Scaler operators in PyTorch. Supported normalizers are L1, L2 and Max.
+    """
+
+    def __init__(self, offset, scale, device):
+        super(Scaler, self).__init__()
+        self.transformer = True
```
interesaaat: You can set this in init.

hummingbird/ml/operator_converters/_scaler_implementations.py
```diff
@@ -0,0 +1,42 @@
...
+
```
interesaaat, line 26: Can you do like self.offset = self.scale = None and then just have one branch each?

interesaaat (approved):

PR #210, opened 2020-07-25
fixing copy-paste error in non-tree onnx tests

ksaur:
This was a copy paste error from our [LGBM](https://github.com/microsoft/hummingbird/blob/master/tests/test_onnxml_lightgbm_converter.py#L42) tests


KranthiGV (approved):
LGTM

PR #231, opened 2020-08-14
Add Sklearn containers for ONNX models

### LINKED ISSUE #233: Add Sklearn API container to ONNX models

interesaaat:
Closes #233.

### DIFF
--- a/hummingbird/ml/_container.py
+++ b/hummingbird/ml/_container.py
@@ -313,13 +313,13 @@ class ONNXSklearnContainer(ABC):
         return named_inputs
 
 
-class ONNXContainerTransformer(ONNXSklearnContainer):
+class ONNXSklearnContainerTransformer(ONNXSklearnContainer):
     """
     Container mirroring Sklearn transformers API.
     """
 
     def __init__(self, model, extra_config={}):
-        super(ONNXContainerTransformer, self).__init__(model, extra_config)
+        super(ONNXSklearnContainerTransformer, self).__init__(model, extra_config)
 
         assert len(self.output_names) == 1
 
@@ -333,13 +333,13 @@ class ONNXContainerTransformer(ONNXSklearnContainer):
         return self.session.run(self.output_names, named_inputs)
 
 
-class ONNXContainerRegression(PyTorchTorchscriptSklearnContainer):
+class ONNXSklearnContainerRegression(ONNXSklearnContainer):
     """
     Container mirroring Sklearn regressors API.
     """
 
     def __init__(self, model, extra_config={}, is_regression=True, is_anomaly_detection=False, **kwargs):
-        super(ONNXContainerRegression, self).__init__(model, extra_config)
+        super(ONNXSklearnContainerRegression, self).__init__(model, extra_config)
 
         assert not (is_regression and is_anomaly_detection)
         if is_regression:
@@ -363,13 +363,13 @@ class ONNXContainerRegression(PyTorchTorchscriptSklearnContainer):
             return self.session.run(self.output_names[0], named_inputs)
 
 
-class ONNXContainerClassification(PyTorchSklearnContainerRegression):
+class ONNXSklearnContainerClassification(ONNXSklearnContainerRegression):
     """
     Container mirroring Sklearn classifiers API.
     """
 
     def __init__(self, model, extra_config={}):
-        super(PyTorchSklearnContainerClassification, self).__init__(model, extra_config, is_regression=False)
+        super(ONNXSklearnContainerClassification, self).__init__(model, extra_config, is_regression=False)
 
         assert len(self.output_names) == 2
 
@@ -383,13 +383,13 @@ class ONNXContainerClassification(PyTorchSklearnContainerRegression):
         return self.session.run(self.output_names[1], named_inputs)
 
 
-class PyTorchSklearnContainerAnomalyDetection(PyTorchSklearnContainerRegression):
+class ONNXSklearnContainerAnomalyDetection(ONNXSklearnContainerRegression):
     """
     Container mirroring Sklearn anomaly detection API.
     """
 
     def __init__(self, model, extra_config={}):
-        super(PyTorchSklearnContainerAnomalyDetection, self).__init__(
+        super(ONNXSklearnContainerAnomalyDetection, self).__init__(
             model, extra_config, is_regression=False, is_anomaly_detection=True
         )
 
--- a/hummingbird/ml/_container.py
+++ b/hummingbird/ml/_container.py
@@ -299,14 +299,19 @@ class ONNXSklearnContainer(ABC):
         if onnx_runtime_installed():
             import onnxruntime as ort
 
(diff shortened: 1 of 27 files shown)


interesaaat:
This PR introduces a breaking change because we are now returning a container rather than a onnx model as result. I added an option to disable the converter, but by default I would like to have containers on because they give us a uniform API. We will have to update the onnxmltools if this PR is merged as is.

@ksaur

tests/test_onnxml_decision_tree_converter.py
```diff
@@ -251,19 +252,20 @@ def test_random_forest_multi_random(self):
...
-        warnings.filterwarnings("ignore")
-        np.random.seed(0)
-        X = np.random.rand(1, 1)
-        X = np.array(X, dtype=np.float32)
-        y = np.random.randint(1, size=1)
-        model = RandomForestClassifier(n_estimators=5).fit(X, y)
-        model.fit(X, y)
-        self._test_classifier(X, model)
+    # # Used for small tree tests
+    # # Commenting this test for the moment because it hits a bug in ORT / ONNXMLTOOLS (https://github.com/onnx/onnxmltools/issues/415)
```
ksaur, line 256: Ah, this is why the coverage drop, right? thanks for filing the[ onnx issue](https://github.com/onnx/onnxmltools/issues/415)!

hummingbird/ml/_container.py
```diff
@@ -275,3 +281,138 @@ def score_samples(self, *inputs):
...
+    The container allows to mirror the Sklearn API.
+    """
+
+    def __init__(self, model, extra_config={}):
+        """
+        Args:
+            model: A ONNX model
+            extra_config: Some additional configuration parameter
+        """
+        if onnx_runtime_installed():
```
ksaur, line 299: i think we need another fail/`else` case here, else it will die on like 305 if this is false

hummingbird/ml/_container.py
```diff
@@ -275,3 +281,138 @@ def score_samples(self, *inputs):
...
+
+        self.model = model
+        self._extra_config = extra_config
+
+        self._session = ort.InferenceSession(self.model.SerializeToString())
+        self._output_names = [self._session.get_outputs()[i].name for i in range(len(self._session.get_outputs()))]
+        self.input_names = [input.name for input in self._session.get_inputs()]
+
+    def _get_named_inputs(self, *inputs):
+        assert len(inputs) == len(self.input_names)
```
ksaur, line 315: can you please add a comment here? (onnx inputs are so confusing!)

hummingbird/ml/_topology.py
```diff
@@ -100,51 +104,91 @@ def convert(topology, backend, device, extra_config={}):
...
+        # Set the device for the model.
         if device != "cpu":
-            test_data.to(device)
-        torch_model = torch.jit.trace(torch_model, test_data).eval()
+            if backend == torch.__name__ or torch.jit.__name__:
+                torch_model = torch_model.to(device)
 
-    if operator_map[operators[-1].full_name].regression:
+        # If the backend is tochscript, jit the model.
+        if backend == torch.jit.__name__:
```
ksaur, line 120: is any of this related to onnx, or is this a separate fix?

hummingbird/ml/_topology.py
```diff
@@ -100,51 +104,91 @@ def convert(topology, backend, device, extra_config={}):
...
+        and not operator_map[operators[idx].full_name].regression
+        and not operator_map[operators[idx].full_name].classification
+        and not operator_map[operators[idx].full_name].anomaly_detection
+        and not operator_map[operators[idx].full_name].transformer
+    ):
+        idx -= 1
+
+    assert idx >= 0, "Cannot detect container type. Please fill an issue at https://github.com/microsoft/hummingbird."
+
+    # If is a transformer, we need to check wheter there is another operator type before.
```
ksaur: s/wheter/whether

I'm a big confused here still...what is going on line 156/157?  This is just "return to the end/original idx" right? so that we don't end up with something negative/nonsensical?

hummingbird/ml/_topology.py
```diff
@@ -100,51 +104,91 @@ def convert(topology, backend, device, extra_config={}):
...
+    if operator_map[operators[idx].full_name].regression:
         # We are doing a regression task.
         if backend == torch.jit.__name__:
             container = TorchScriptSklearnContainerRegression
+        elif backend == onnx.__name__:
+            container = ONNXSklearnContainerRegression
         else:
             container = PyTorchSklearnContainerRegression
-    elif operator_map[operators[-1].full_name].anomaly_detection:
+    elif operator_map[operators[idx].full_name].anomaly_detection:
```
ksaur, line 167: And this is guaranteed to always be in-range because of 157?

hummingbird/ml/operator_converters/sklearn/skl_sv.py
```diff
@@ -18,7 +18,7 @@
 
 class SVC(BaseOperator, torch.nn.Module):
     def __init__(self, kernel, degree, sv, nv, a, b, gamma, coef0, classes, device):
-        super(SVC, self).__init__()
+        super(SVC, self).__init__(classification=True)
```
ksaur, line 21: this is due to the container change, right? do we need to change this in other places, or is this sort of a bugfix?

tests/test_lightgbm_converter.py
```diff
@@ -280,10 +280,7 @@ def test_lightgbm_onnx(self):
         onnx_model = hummingbird.ml.convert(model, "onnx", X)
 
         # Get the predictions for the ONNX-ML model
-        session = ort.InferenceSession(onnx_model.SerializeToString())
-        output_names = [session.get_outputs()[i].name for i in range(len(session.get_outputs()))]
-        inputs = {session.get_inputs()[0].name: X}
-        onnx_pred = session.run(output_names, inputs)
+        onnx_pred = onnx_model.predict(X)
```
ksaur, line 283: ok, i understand now about the breaking change.  i think this is great!  (well, not the breaking part, but much more usable.)

tests/test_onnxml_decision_tree_converter.py
interesaaat: This plus some corner conditions I think.

hummingbird/ml/operator_converters/sklearn/skl_sv.py
interesaaat: I think it needs to go in all classification cases. Hopefully I got all of them. For trees I set the type after the class initialization (because as the trees are implemented we cannot do at instance creation time).

tests/test_lightgbm_converter.py
interesaaat: Yea that is what I tought too!

hummingbird/ml/_container.py
interesaaat: Yes I definitely forgot this one.

PR #246, opened 2020-08-21
Strip torch

interesaaat:
This PR set a generic torch target so that we don't need to support different versions during release. 

The only drawback is that going forward we expect users to manually download PyTorch. This is not necessarily a bad thing since before the workflow for GPU was weird (i.e., install HB, then remove PyTorch and then reinstall PyTorch with CUDA).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,14 +30,13 @@ For more information see the [Code of Conduct FAQ](https://opensource.microsoft.
 contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
 
 ## Developing
-The simplest setup is:
+The simplest setup (assuming you have PyTorch already installed) is:
 ```
 mkdir hummingbird
 cd hummingbird
 git clone https://github.com/microsoft/hummingbird.git .
 pip install -e .[docs,tests,extra]
 ```
-On Windows, the last line above must also contain `-f https://download.pytorch.org/whl/torch_stable.html`. (This is required because PyTorch version for Windows is not up to date.)
 
 ### Docker
 We provide a simple [Dockerfile](https://github.com/microsoft/hummingbird/blob/master/Dockerfile) that you can customize to your preferred development environment.
@@ -45,7 +44,7 @@ We provide a simple [Dockerfile](https://github.com/microsoft/hummingbird/blob/m
 docker build git://github.com/microsoft/hummingbird -t hb-jupy
 docker run -it hb-dev
 ```
-### Codespases
+### Codespaces
 For a light-weight, web-based experience, we provide the configuration ([.devcontainer](https://github.com/microsoft/hummingbird/tree/master/.devcontainer)) for [Codespaces](https://online.visualstudio.com/environments).  More information on this setup can be found [here]( https://docs.microsoft.com/en-us/visualstudio/online/reference/configuring).
 
 ### Tools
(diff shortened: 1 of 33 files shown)


interesaaat:
We got a problem with the ci/cd pipeline. Hold on.

ksaur:
Oddly enough, there is an "if" but not "else" available in GH Actions
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif

ksaur:
Windows with python 3.5 is such a problem child.

maybe we can try https://stackoverflow.com/questions/61488902/cannot-import-pytorch-winerror-126-the-specified-module-could-not-be-found ?

CONTRIBUTING.md
```diff
@@ -30,22 +30,23 @@ For more information see the [Code of Conduct FAQ](https://opensource.microsoft.
...
+The simplest setup (assuming you have PyTorch already installed) is:
 ```
 mkdir hummingbird
 cd hummingbird
 git clone https://github.com/microsoft/hummingbird.git .
 pip install -e .[docs,tests,extra]
 ```
-On Windows, the last line above must also contain `-f https://download.pytorch.org/whl/torch_stable.html`. (This is required because PyTorch version for Windows is not up to date.)
+
+If you do not have pytorch installed, please [install it](https://pytorch.org/get-started/locally/) first.
```
ksaur, line 41: it seems `pip install torch==1.6.* -f https://download.pytorch.org/whl/torch_stable.html` should work in  MOST cases...do you want to suggest that here?  I'm fine  with also keeping this as a simple link :)

tests/test_onnxml_lightgbm_converter.py
```diff
@@ -78,6 +78,7 @@ def _test_classifier(self, X, model, rtol=1e-06, atol=1e-06, extra_config={}):
     @unittest.skipIf(
         not (onnx_ml_tools_installed() and onnx_runtime_installed()), reason="ONNXML test require ONNX, ORT and ONNXMLTOOLS"
     )
+    @unittest.skipIf(not lightgbm_installed(), reason="LightGBM test requires LightGBM installed")
```
ksaur, line 81: how was this working before??

CONTRIBUTING.md
interesaaat: It won't because you need to figure which cuda version to use if you want to go with gpu. Like +cu101. I think it is better to tell to just check the website.

tests/test_onnxml_lightgbm_converter.py
interesaaat: Right? I thought the same!

setup.py
```diff
@@ -21,14 +21,7 @@
...
 
-install_requires = ["numpy>=1.15", "onnxconverter-common>=1.6.0", "scikit-learn==0.22.1"]
-if sys.platform == "darwin":
-    install_requires.append("torch")
-else:
-    if sys.version_info[1] == 5:
-        install_requires.append("torch==1.5.1+cpu")
-    else:
-        install_requires.append("torch==1.6.0+cpu")
+install_requires = ["numpy>=1.15", "onnxconverter-common>=1.6.0", "scikit-learn>=0.22.1", "torch>=1.4.0"]
```
ksaur: Do we still want `torch` here?

interesaaat: I think so, otherwise HB will crash with "torch is not installed". The >= 1.4.0 should cover all the "cpu", "cu" etc. versions. Actually, maybe >= 1.4.* is better.

ksaur: ok, makes sense.

OH! is there a way to make a nice "please install torch" message?

interesaaat: We have something like ""To use Hummingbird you need to install torch.""

ksaur: ah perfect, now i see `assert torch_installed(), "To use Hummingbird you need to install torch."`

PR #251, opened 2020-08-26
Add basic support for scikit-learn pipeline

### LINKED ISSUE #250: Add support for scikit-learn pipelines
For the moment Hummingbird is only able to support single model \ featurizers.

We should also add support for `Pipeline`objects, as well as `FutureUnion` and `ColumnTransform`.

interesaaat:
This PR adds basic support for scikit-learn `Pipeline`, `FeatureUnion` and `ColumnTransformer`. It covers the basic cases, while we need more testing for corner cases.

I also realize that we need to support pandas data frames as input format and automatically extract schema information from it. I will open a related issue.

Closes #250.


hummingbird/ml/_container.py
```diff
@@ -52,24 +52,57 @@ def __init__(self, input_names, output_names, operator_map, operators, extra_con
...
-        self._output_names = output_names
+
+        # Define input \ output names.
+        # This is required because the internal variable names may differ from the original (raw) one.
+        # This may happen, for instance, because we force our internal naming to be unique.
+        self._input_names = []
+        self._output_names = []
+        map = {}
+        for op in operators:
+            for op_input in op.inputs:
```
ksaur: i understand why the first two loops, but is the third loop always necessary?  Is it possible to do the break (line 67) earlier?

hummingbird/ml/_container.py
```diff
@@ -52,24 +52,57 @@ def __init__(self, input_names, output_names, operator_map, operators, extra_con
...
+                for i_name in input_names:
+                    if op_input.raw_name == i_name and i_name not in map:
+                        map[op_input.raw_name] = op_input.full_name
+            if len(input_names) == 0:
+                break
+        for i_name in input_names:
+            self._input_names.append(map[i_name])
+
+        map = {}
```
ksaur: lines 61-70 are almost identical to 72-81. is there any way to combine?

hummingbird/ml/_parse.py
```diff
@@ -19,6 +21,13 @@
 from ._utils import sklearn_installed
 from .supported import get_sklearn_api_operator_name, get_onnxml_api_operator_name
 
+if sklearn_installed():
```
ksaur: Is this ever false since it's a requirement of the install for HB?

hummingbird/ml/_parse.py
```diff
@@ -162,33 +171,128 @@ def _parse_sklearn_single_model(scope, model, inputs):
...
+    :return: A list of output variables produced by the input pipeline
     """
     for step in model.steps:
         inputs = _parse_sklearn_api(scope, step[1], inputs)
     return inputs
 
 
-def _build_sklearn_api_parsers_map():
-    from sklearn import pipeline
+def _parse_sklearn_feature_union(scope, model, inputs):
```
ksaur, line 185: add link to https://github.com/onnx/sklearn-onnx/blob/master/skl2onnx/_parse.py

hummingbird/ml/operator_converters/sklearn/array_feature_extractor.py
```diff
@@ -53,5 +53,24 @@ def convert_sklearn_variance_threshold(operator, device, extra_config):
     return ArrayFeatureExtractor(np.ascontiguousarray(indices), device)
 
 
+def convert_sklearn_array_feature_extractor(operator, device, extra_config):
+    """
+    Converter for `sklearn.feature_selection.VarianceThreshold`.
```
ksaur: s/VarianceThreshold/AFE/g  (also  line 61)

hummingbird/ml/operator_converters/sklearn/linear.py
```diff
@@ -30,7 +30,7 @@ def convert_sklearn_linear_model(operator, device, extra_config):
     """
     classes = [0] if not hasattr(operator.raw_operator, "classes_") else operator.raw_operator.classes_
 
-    if not all([type(x) in [int, np.int32, np.int64] for x in classes]):
+    if not all(["int" in str(type(x)) for x in classes]):
```
ksaur, line 33: was this a bug,  or related to pipelines?  (if not, maybe make a separate PR next time for refactorings to avoid confusion?)

hummingbird/ml/operator_converters/sklearn/pipeline.py
```diff
@@ -0,0 +1,74 @@
...
+
+import numpy as np
+from onnxconverter_common.registration import register_converter
+import torch
+
+from .. import constants
+from .._base_operator import BaseOperator
+
+
+class Concat(BaseOperator, torch.nn.Module):
```
ksaur, line 20: I see in `supported.py` you have a comment on "These are ops injected by the parser not "real" sklearn operators."

can you add some other  comment  about that on these two? otherwise they look like new features

hummingbird/ml/_parse.py
```diff
@@ -283,6 +385,77 @@ def _remove_zipmap(node_list):
...
+    """
+    pt_var = None
+    pt_is = []
+    for p in indices:
+        pt_v, pt_i = _get_column_index(p, inputs)
+        pt_is.append(pt_i)
+        if pt_var is None:
+            pt_var = pt_v
+        elif pt_var != pt_v:
+            cols = [pt_var, pt_v]
```
ksaur: i  think line 450 is not necessary because we err

hummingbird/ml/_container.py
interesaaat: I tried but I don't it is possible.

hummingbird/ml/_container.py
interesaaat: Yep, fixed.

hummingbird/ml/_parse.py
interesaaat: yep, let me remove it. (BTW do you remember why sklearn has always to be installed?)

hummingbird/ml/_parse.py
interesaaat: Fixed.

hummingbird/ml/operator_converters/sklearn/array_feature_extractor.py
interesaaat: Fixed.

hummingbird/ml/operator_converters/sklearn/linear.py
interesaaat: This was something crashing some pipeline test on win machines (no idea why TBH).

hummingbird/ml/operator_converters/sklearn/pipeline.py
interesaaat: yep.

hummingbird/ml/_parse.py
interesaaat: Yea I thought about that too. Let me fix it.

hummingbird/ml/_parse.py
ksaur: we tried to remove it, but then we realized that xgb,lgbm,etc all take dependencies on it.
Without at least [skl | onnx] installed, HB is unusable, so we decided to leave it

hummingbird/ml/_parse.py
interesaaat: Ah but lgbm and xgb are extra now, no? So in theory it should work without all of them (I think).

ksaur: "work" but do nothing

PR #253, opened 2020-08-27
Bump xgb version

### LINKED ISSUE #2: Fix xgb converter for version > 0.90
Regression for XGBoost is failing for 1.0.2. 

Apparently there was:
- a change in the API, ``alpha`` is now an array\list
- a change in the model structure because we are getting different results for regression
- a change in ``max_depth`` because ``None`` does not work as ``max_depth`` in 0.90

interesaaat:
We were forcing ourselves to `xgb==0.90`. This PR enable later xgb versions.

### DIFF
--- a/doc/html/hummingbird/ml/operator_converters/xgb.html
+++ b/doc/html/hummingbird/ml/operator_converters/xgb.html
@@ -27,7 +27,7 @@
 <details class="source">
 <summary>
 <span>Expand source code</span>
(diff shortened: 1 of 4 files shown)


ksaur:
is `base_prediction` the only change we need? I thought there was some problem with a type change with `alpha`, but maybe that is for operators we haven't ported in yet?

interesaaat:
> is `base_prediction` the only change we need? I thought there was some problem with a type change with `alpha`, but maybe that is for operators we haven't ported in yet?

Yes that's the only change. We renamed `alpha` into `base_prediction` when we refactored the trees (I think the term is more generic, it was not clear what "alpha" means).

PR #255, opened 2020-08-27
DOC: Cleanup remnants from copied pdoc template

kernc:
Strongly suggested to apply this one way or another before rebuilding API docs. :grin:


interesaaat:
Thanks @kernc for the contribution! This is definitely something that we missed :see_no_evil:

PR #258, opened 2020-08-28
Adding support for sklearn.preprocessing.Binarizer

ksaur:


hummingbird/ml/operator_converters/_binarizer_implementations.py
```diff
@@ -0,0 +1,27 @@
...
+    Class implementing Binarizer operators in PyTorch.
+    """
+
+    def __init__(self, threshold, device):
+        super(Binarizer, self).__init__()
+        self.transformer = True
+        self.threshold = torch.nn.Parameter(torch.FloatTensor([threshold]), requires_grad=False)
+
+    def forward(self, x):
+        return torch.gt(x, self.threshold).float()
```
interesaaat: Do we actually need cast to float?

hummingbird/ml/operator_converters/_binarizer_implementations.py
```diff
@@ -0,0 +1,27 @@
+# -------------------------------------------------------------------------
```
interesaaat: Usually we put into `_*_implementations.py` all the implementations shared between onnx and pytorch. In this case we don't have an onnx implementation, so I think we can just have the implementation into the `binarizer.py` file.

ksaur: I considered this, but do we have an example currently that is just in a .py?  Not for SKL.  Maybe a few for onnx?

interesaaat: Yes. Like `sv` and `iforest`.

PR #260, opened 2020-08-29
Issue #153 - Add SKL Other Classifiers

### LINKED ISSUE #153: Add SKL Other Classifiers
Add support for other SKL classifiers (i.e., other than linear models and tree-based models)

This includes:

- MLPClassifier
- BernoulliNB
- GaussianNB
- MultinomialNB

scnakandala:
This pull request adds Hummingbird support for the following Scikit Learn Models

1. MLPClassifer
2. BernoulliNB
3. GaussianNB
4. MultinomialNB

Testing:
- Added a new test file named test_sklearn_mlp_converter.py to test MLPClassifer conversion
- Added a new test file named test_sklearn_nb_converter.py to test the conversion of BernoulliNB, GaussianNB, and MultinomialNB models.


interesaaat:
LGTM. One minor thing if you can: we put implementations into `_implementation.py` files only if it's shared among other backend (e.g., with ONNX) otherwise we leave the implementations into the convert files. (So basically if you can get rid of the implementation files and just move the classes into the converter files).

This is second time in 2 days that I noticed this, do you think that is better to put all implementations into `_implementation.py` files or in a implementation directory? @ksaur

ksaur:
I think as long as we are consistent.  I  pulled it out of the `_impl` at your suggestion i think  last time...exactly  like this case. :-D

ksaur:
What will make the most sense as we merge in more frameworks?
Your call @interesaaat   :)

interesaaat:
Since both of you put the implementations into separate files I think that is more natural at this point.

scnakandala:
Is there a concern about how we generate automatic documentation?
Do we exclude doc blocks from certain directories?

interesaaat:
Nope, pdoc should generate documentation for all files minus the one starting with '_'. But we override this on the main `__init__` file.

hummingbird/ml/operator_converters/__init__.py
```diff
@@ -30,6 +30,8 @@
 from .sklearn import gbdt  # noqa: E402
 from .sklearn import iforest  # noqa: E402
 from .sklearn import linear as sklearn_linear  # noqa: E402
+from .sklearn import mlp as sklearn_mlp  # noqa: E402
+from .sklearn import nb as sklearn_nb  # noqa: E402
```
ksaur, line 34: But now that we are keeping the _mlp_implementations.py and _nb_implementations.py , we  need to add that to lines 48 of this file for pdoc to work. :)

ksaur: (other than that i think we are ready to merge!   :) )

scnakandala: Ok. I will change and update the pull request.

PR #261, opened 2020-08-29
Issue #254 - LogisticRegression returns wrong results with lbfgs

### LINKED ISSUE #254: LogisticRegression returns wrong results with lbfgs
With `lbfgs` solver the results are wrong. To replicate:

```
model = LogisticRegression(C=0.01, class_weight=dict(zip([False, True], [0.2, 0.8])), n_jobs=1, max_iter=10, solver="lbfgs", tol=1e-3,)
        
np.random.seed(0)
X = np.random.rand(100, 200)
X = np.array(X, dtype=np.float32)
y = np.random.randint(2, size=100)

model.fit(X, y)

torch_model = hummingbird.ml.convert(model, "torch")

self.assertTrue(torch_model is not None)
np.testing.assert_allclose(model.predict_proba(X), torch_model.predict_proba(X), rtol=1e-6, atol=1e-6)

```

scnakandala:
This pull request fixes the issue in Logistic Regression with auto multiclass option and binary data.
Quoting the corresponding section from sklearn documentation:

multi_class{‘auto’, ‘ovr’, ‘multinomial’}, default=’auto’
If the option chosen is ‘ovr’, then a binary problem is fit for each label. For ‘multinomial’ the loss minimised is the multinomial loss fit across the entire probability distribution, even when the data is binary. ‘multinomial’ is unavailable when solver=’liblinear’. ‘auto’ selects ‘ovr’ if the data is binary, or if solver=’liblinear’, and otherwise selects ‘multinomial’.

Testing:
- Extended the existing LogisticRegression test file to include test cases with multiclass option set to auto and binary classification.

### DIFF
--- a/doc/html/hummingbird/ml/operator_converters/sklearn/linear.html
+++ b/doc/html/hummingbird/ml/operator_converters/sklearn/linear.html
@@ -26,7 +26,7 @@
 <details class="source">
 <summary>
 <span>Expand source code</span>
-<a href="https://github.com/microsoft/hummingbird/blob/master/hummingbird/ml/operator_converters/sklearn/linear.py#L0-L73" class="git-link">Browse git</a>
+<a href="https://github.com/microsoft/hummingbird/blob/master/hummingbird/ml/operator_converters/sklearn/linear.py#L0-L75" class="git-link">Browse git</a>
 </summary>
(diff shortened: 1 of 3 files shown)


ksaur:
Wow, nice fix! Thanks for this!! :)

PR #262, opened 2020-08-29
Issue #200: Adding SelectKBest chi2 support

### LINKED ISSUE #200: SelectKBest does not work with chi2
Our tests fail when we use `chi2` as the function with `SelectKBest` (Ex:`SelectKBest(chi2, k=20)`).  There is some additional implementation needed for this functionality.

`SelectKBest` does work with some functions such as `mutual_info_classif`   (See PR #198  )

scnakandala:
This pull request adds support for SelectKBest with chi2 support.

Testing:
- Uncommented the chi2 test which was previously commented out.


ksaur:
Thanks @scnakandala  !!  I was stuck on one!

PR #265, opened 2020-08-29
Fix bug with empty trees and multiclass

interesaaat:


ksaur:
I think that codecov is doing some weird stuff.  How is it possible that adding these 2 lines tanks coverage by almost 1%??

Is codecov broken as suspected, or am I missing something?   (clicking into the codecov webfiles is just broken.)

interesaaat:
> I think that codecov is doing some weird stuff. How is it possible that adding these 2 lines tanks coverage by almost 1%??
> 
> Is codecov broken as suspected, or am I missing something? (clicking into the codecov webfiles is just broken.)

Yea I noticed that too. This should not change the coverage at all!

ksaur:
Okie dokie, let's merge and hope the coverage doesn't puke.
else we can host a "coverage recovery" afternoon. :-D

PR #266, opened 2020-08-29
Add float to binarizer output

interesaaat:
(Sorry Karla, you were right)


ksaur:
What bug did you hit? Do we need to add another test?

interesaaat:
> What bug did you hit? Do we need to add another test?

I dunno. I just run it into a pipeline and the successive operator expected `float` (which actually makes sense).

ksaur:
> I dunno. I just run it into a pipeline and the successive operator expected `float` (which actually makes sense).

Ah ok!  I have been trying very hard to break this as-written with a test...but we are not testing expected output by successive operators, so that makes sense. Maybe we should add that in the pipeline testing!

PR #267, opened 2020-08-29
add sklearn.impute.SimpleImputer

ksaur:


interesaaat:
LGTM but since `MissingIndicator` and `SimpleImputer` are using to impute missing values, I think we can just have a single file (e.g., `impute.py`) containing both of them. WDYT?

PR #268, opened 2020-08-29
sklearn.impute.MissingIndicator converter

ksaur:
MissingIndicator converter needs a bugfix and several additional tests to hit all the cases.

### DIFF
--- a/hummingbird/ml/operator_converters/sklearn/missing_indicator.py
+++ b/hummingbird/ml/operator_converters/sklearn/missing_indicator.py
@@ -1,0 +1,57 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+"""
+Converter for scikit-learn MissingIndicator.
+"""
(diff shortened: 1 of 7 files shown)


hummingbird/ml/operator_converters/sklearn/missing_indicator.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    def __init__(self, sklearn_missing_indicator, device):
+        super(MissingIndicator, self).__init__()
+        self.transformer = True
+        self.missing_values = torch.nn.Parameter(
+            torch.tensor([sklearn_missing_indicator.missing_values], dtype=torch.float32), requires_grad=False
+        )
+        self.features = sklearn_missing_indicator.features
+        self.is_nan = True if (self.missing_values in ["NaN", None, np.nan]) else False
```
scnakandala: @ksaur Can you try changing this to
`self.is_nan = True if (sklearn_missing_indicator.missing_values in ["NaN", None, np.nan]) else False`

ksaur: Ah!! Great catch!!! THANKS.

(You must be some kind of wizard think.)

PR #269, opened 2020-08-30
sklearn.preprocessing.PolynomialFeatures

ksaur:
Currently this supports only degree 2, and does not support interaction_only


interesaaat:
@scnakandala any idea what is going on with `interaction_only`?

scnakandala:
Technically it is possible. I think I wanted to prototype it quickly so I restricted the configurations to defaults.
I can have a look at this.

interesaaat:
> Technically it is possible. I think I wanted to prototype it quickly so I restricted the configurations to defaults.
> I can have a look at this.

Ah ok. Don't worry. I thought it was more like a bug. This is not super important, we can merge as is and open an issue.

ksaur:
Rebased, issue filed, ready to merge. :)

Hemantr05:
Does #269 still require fixing?

PR #274, opened 2020-08-31
Add imputer for sklearn 0.21.3

interesaaat:
For some reason isolation forest now returns some wrong prediction for 0.21.3. I have to check what changed and fix it. I will open a separate issue on this.

### DIFF
--- a/hummingbird/ml/operator_converters/sklearn/imputer.py
+++ b/hummingbird/ml/operator_converters/sklearn/imputer.py
@@ -28,10 +28,12 @@ class SimpleImputer(BaseOperator, torch.nn.Module):
         self.replace_values = torch.nn.Parameter(
             torch.tensor([sklearn_imputer.statistics_], dtype=torch.float32), requires_grad=False
         )
-        self.missing_values = torch.nn.Parameter(
-            torch.tensor([sklearn_imputer.missing_values], dtype=torch.float32), requires_grad=False
-        )
+
         self.is_nan = True if (sklearn_imputer.missing_values == "NaN" or np.isnan(sklearn_imputer.missing_values)) else False
+        if not self.is_nan:
+            self.missing_values = torch.nn.Parameter(
+                torch.tensor([sklearn_imputer.missing_values], dtype=torch.float32), requires_grad=False
+            )
 
     def forward(self, x):
         if self.is_nan:
@@ -99,5 +101,6 @@ def convert_sklearn_missing_indicator(operator, device, extra_config):
     return MissingIndicator(operator.raw_operator, device)
 
 
+register_converter("SklearnImputer", convert_sklearn_simple_imputer)
 register_converter("SklearnSimpleImputer", convert_sklearn_simple_imputer)
 register_converter("SklearnMissingIndicator", convert_sklearn_missing_indicator)
--- a/hummingbird/ml/supported.py
+++ b/hummingbird/ml/supported.py
@@ -174,6 +174,7 @@ def _build_sklearn_operator_list():
             VarianceThreshold,
         ]
 
+        # Remove all deprecated operators given the sklearn version. E.g., Imputer for sklearn > 0.21.3.
         return [x for x in supported_ops if x is not None]
 
     return []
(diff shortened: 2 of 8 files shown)


hummingbird/ml/operator_converters/sklearn/imputer.py
```diff
@@ -99,5 +101,6 @@ def convert_sklearn_missing_indicator(operator, device, extra_config):
     return MissingIndicator(operator.raw_operator, device)
 
 
+register_converter("SklearnImputer", convert_sklearn_simple_imputer)
```
ksaur, line 104: Is this only necessary for backward compatibility reasons?

hummingbird/ml/operator_converters/sklearn/imputer.py
```diff
@@ -28,10 +28,12 @@ def __init__(self, sklearn_imputer, device):
         self.replace_values = torch.nn.Parameter(
             torch.tensor([sklearn_imputer.statistics_], dtype=torch.float32), requires_grad=False
         )
-        self.missing_values = torch.nn.Parameter(
-            torch.tensor([sklearn_imputer.missing_values], dtype=torch.float32), requires_grad=False
-        )
+
         self.is_nan = True if (sklearn_imputer.missing_values == "NaN" or np.isnan(sklearn_imputer.missing_values)) else False
+        if not self.is_nan:
```
ksaur, line 33: Can you add a comment here about the bug requiring this order?

hummingbird/ml/supported.py
```diff
@@ -119,10 +119,16 @@ def _build_sklearn_operator_list():
             StandardScaler,
         )
 
+        try:
+            from sklearn.preprocessing import Imputer
+        except ImportError:
+            # Imputer was deprecate in sklearn >= 0.22
```
ksaur, line 125: We could also do something like 
```
import sklearn
if sklearn.__version__ < 0.22:
....
```
but maybe this way is simpler :)

hummingbird/ml/supported.py
```diff
@@ -167,6 +174,8 @@ def _build_sklearn_operator_list():
             VarianceThreshold,
         ]
 
+        return [x for x in supported_ops if x is not None]
```
ksaur, line 178: Was there some bug?

setup.py
```diff
@@ -21,7 +21,7 @@
     if start_pos >= 0:
         long_description = long_description[start_pos:]
 
-install_requires = ["numpy>=1.15", "onnxconverter-common>=1.6.0", "scikit-learn>=0.22.1", "torch>=1.4.*"]
+install_requires = ["numpy>=1.15", "onnxconverter-common>=1.6.0", "scikit-learn>=0.21.3", "torch>=1.4.*"]
```
ksaur, line 24: So this is necessary for pipelines?

hummingbird/ml/operator_converters/sklearn/imputer.py
interesaaat: Yes. Because Imputer was deprecated in 0.22.

hummingbird/ml/supported.py
interesaaat: Yes. That is what I was doing at the beginning but then realized that this is way simpler! :)

setup.py
interesaaat: Yes. To get Imputer in.

hummingbird/ml/supported.py
interesaaat: Imputer is None if sklearn >= 0.22. I can add a comment.

hummingbird/ml/operator_converters/sklearn/imputer.py
interesaaat: I had to switch the 2 because when `self.is_nan` is true, `self.missing_values = torch.nn.Parameter(torch.tensor([sklearn_imputer.missing_values], dtype=torch.float32), requires_grad=False)` creates a tensor with `nan` as value which crashed PyTorch. (No idea how it was actually working before).

PR #290, opened 2020-09-05
fix isolation forest for sklearn <= 0.21

### LINKED ISSUE #275: Isolation Forest results do not match in sklearn 0.21.3
To reproduce: run isolation forest tests with `scikit-learn==0.21.3`

zhanjiezhu:
fix #275 

[the predict function of sklearn <= 0.21](https://github.com/scikit-learn/scikit-learn/blob/0.21.X/sklearn/ensemble/iforest.py#L316) has a dependency on an extra attribute "threshold_", which was removed after 0.21. To make it compatible, parse this as constant if it exists otherwise set 0. Then change accordingly in the functions e.g. decision_function. Note that it's not possible to combine this value into the existing "offset" because only predict relies on "threshold", but decision_function & score_samples do not. Should works for sklearn >= 0.20

### DIFF
--- a/doc/html/hummingbird/ml/operator_converters/constants.html
+++ b/doc/html/hummingbird/ml/operator_converters/constants.html
@@ -26,7 +26,7 @@
 <details class="source">
 <summary>
 <span>Expand source code</span>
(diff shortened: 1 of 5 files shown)


interesaaat:
Thanks @zhanjiezhu! I fixed a couple of things.

PR #295, opened 2020-09-13
Checking for strings in pipeline col index

ksaur:
The "else" clause of `_get_column_index` (strings) needs to be commented out.  It succeeds at parsing the columns, but then throws an error in `parse_sklearn_api_model` for `np.str`. I added a test for the fail case, which we can flip to a regular test later :)


interesaaat:
Nice catch! Do we have an issue for this?

ksaur:
My goal was to bump coverage up a bit, which is when i found this. :)

I don't think there is an issue, i will make it!

Issue #252, opened 2020-08-26
Add support for Pandas Dataframes


interesaaat:
In Hummingbird we currently support `ndarrays` as input. We should add support for Pandas `Dataframes` too. This is particularly critical for pipelines since each operator in a pipelines can work on a set of columns on the dataframe. This will require extracting schema information from the dataframe.

scnakandala:
> each operator in a pipelines can work on a set of columns on the dataframe

@interesaaat  Will the columns be referenced by the index or by the column name?
I remember there are some operators like ColumnTranformer which refer columns by index.
I think these cases should be easy to support by simply converting the dataframe to numpy (df.to_numpy())

interesaaat:
Yes this is exactly what we have right now. It would be good to extend it to support dataframes more in general.

PR #300, opened 2020-09-16
Add support for pandas dataframes both at conversion time and inference time

### LINKED ISSUE #297: Strings for pipelines
We need to implement Strings for Pipelines.

When we do, we can change the test (`test_pipeline_column_transformer_string`) in PR #295 to call `convert` rather than check for `NotImplementedError`

interesaaat:
Closes #297 and #252. Still need some work so support strings into Pandas dataframes.


hummingbird/ml/_container.py
```diff
@@ -82,8 +82,19 @@ def _fix_var_naming(operators, names, mod="input"):
 
     def forward(self, *inputs):
         with torch.no_grad():
-            assert len(self._input_names) == len(inputs)
-
+            assert len(self._input_names) == len(inputs) or (
+                hasattr(inputs[0], "columns") and len(self._input_names) == len(inputs[0].columns)
+            ), "numer of inputs or number of collumns in the dataframe do not match with the expected number of inputs {}".format(
```
ksaur: s/numer/number
s/collumns/columns

hummingbird/ml/_container.py
```diff
@@ -82,8 +82,19 @@ def _fix_var_naming(operators, names, mod="input"):
...
+                self._input_names
+            )
+
+            if hasattr(inputs[0], "columns"):  # This is a pandas dataframe
```
ksaur: Is there a better way to check this?

hummingbird/ml/convert.py
```diff
@@ -149,14 +149,18 @@ def _convert_onnxml(model, backend, test_input, device, extra_config={}):
...
             assert len(initial_types[0][1].shape) == 2, "Hummingbird currently support only inputs with len(shape) == 2."
 
-            from onnxconverter_common.data_types import FloatTensorType, Int32TensorType
+            from onnxconverter_common.data_types import FloatTensorType, DoubleTensorType, Int32TensorType, Int64TensorType
 
             test_input = np.random.rand(initial_types[0][1].shape[0], initial_types[0][1].shape[1])
             extra_config[constants.N_FEATURES] = initial_types[0][1].shape[1]
             if type(initial_types[0][1]) is FloatTensorType:
                 test_input = np.array(test_input, dtype=np.float32)
+            elif type(initial_types[0][1]) is DoubleTensorType:
```
ksaur, line 158: please add a test case for this and lines 162.

(Trying to keep coverage above 92% if possible!)

hummingbird/ml/_parse.py
```diff
@@ -514,12 +514,12 @@ def _get_column_index(i, inputs):
...
+        for ind, inp in enumerate(inputs):
+            if inp.onnx_name == i:
+                return ind, 0
+        raise RuntimeError("Unable to find column name '{0}'".format(i))
     else:
-        # TODO: implement strings
-        # for ind, inp in enumerate(inputs):
-        #    if inp.onnx_name == i:
-        #        return ind, 0
-        # raise RuntimeError("Unable to find column name '{0}'".format(i))
```
ksaur, line 522: You can make line 517 be "else" and then delete line 523-525.  there are only two cases.  At least based on [skl2onnx](https://github.com/onnx/sklearn-onnx/blob/9939c089a467676f4ffe9f3cb91098c4841f89d8/skl2onnx/common/utils.py#L102)

hummingbird/ml/operator_converters/_one_hot_encoder_implementations.py
```diff
@@ -78,11 +78,25 @@ def __init__(self, categories, device):
...
-        for i in range(self.num_columns):
-            encoded_tensors.append(torch.eq(x[:, i : i + 1], self.condition_tensors[i]))
+
+        if len(x) > 1:
+            assert len(x) == self.num_columns
+
+            for i in range(self.num_columns):
+                input = x[i]
+                if input.dtype != torch.int64:
+                    input = input.long()
```
ksaur, line 90: can we add a test for this too? (line 90)

tests/test_sklearn_pipeline.py
```diff
@@ -167,39 +157,9 @@ def test_pipeline_column_transformer_string(self):
...
-            "parch",
-            "sibsp",
-            "cabin",
-            "ticket",
-            "name",
-            "body",
-            "home.dest",
-            "boat",
-        }
+        to_drop = {"parch", "sibsp", "cabin", "ticket", "name", "body", "home.dest", "boat", "sex", "embarked"}
```
ksaur, line 162: ooh way easier. smart. 👍

ksaur (commented):
Thanks for getting this out so quickly!

hummingbird/ml/_container.py
interesaaat: good catch!

hummingbird/ml/_container.py
interesaaat: I can add a `pandas_installed()` check but I think it will have a hit in performance (since you load a library at runtime). Maybe I have a better idea, let me take a look.

hummingbird/ml/convert.py
interesaaat: Added.

hummingbird/ml/_parse.py
interesaaat: fixed.

tests/test_sklearn_pipeline.py
interesaaat: yup

hummingbird/ml/operator_converters/_one_hot_encoder_implementations.py
interesaaat: Weird we should already have tests for this

PR #310, opened 2020-09-27
Adding support for SparkML

### LINKED ISSUE #299: Integrate with SparkML
Since there is a pyspark API: https://spark.apache.org/docs/latest/api/python/pyspark.ml.html 

We should be able to embed into HB the sparkml parser as we did with sklearn parser.

scnakandala:
closes #299


scnakandala:
@interesaaat @ksaur I think this pull request is ready for review. I focussed on getting the end-to-end flow working. If this looks good, maybe we can merge this and have separate issues for each operator or operator groups.

ksaur:
Hi @scnakandala, thanks so much for this!  I'm very excited for this feature! I'll take a look tomorrow! :)

In the meantime, I see the coverage is dropping a bit on this. Codecov can be a bit flakey so maybe it's lying, but can you please take a look (particularly at convert.py [here](https://codecov.io/gh/microsoft/hummingbird/pull/310/tree?path=hummingbird%2Fml)) and see if it looks like you expect? If coverage looks right on your machine, it might just be a transient codecov  issue.

Sounds great on having separate issues for additional operator groups!

scnakandala:
@ksaur I think the issue arises because of missing pyspark and/or pandas in the testing environment which causes the Spark tests to be skipped. Will it be possible to add these libraries to the CI setup.

BTW, I added sparkml as an additional directive in the setup.py to install pyspark.
https://github.com/microsoft/hummingbird/pull/310/files?file-filters%5B%5D=.py#diff-2eeaed663bd0d25b7e608891384b7298

hummingbird/ml/_parse.py
```diff
@@ -407,6 +484,62 @@ def _parse_onnx_single_operator(scope, operator):
...
+    # corresponding operator conversion time. For models (pyspark.ml.classification.*, pyspark.ml.regression.*)
+    # we assume there will be only one output.
+    if inspect.getmodule(operator).__name__ == "pyspark.ml.feature":
+        for input in this_operator.inputs:
+            variable = scope.declare_local_variable(input.raw_name)
+            this_operator.outputs.append(variable)
+
+        variable = scope.declare_local_variable("variable")
+        this_operator.outputs.append(variable)
+    else:
```
interesaaat, line 544: These branch can be removed if you remove indentation from the previous 2 lines.

hummingbird/ml/convert.py
```diff
@@ -183,10 +195,30 @@ def _convert_onnxml(model, backend, test_input, device, extra_config={}):
...
 
+def _convert_sparkml(model, backend, test_input, device, extra_config={}):
+    """
+    This function converts the specified *Spark-ML* (API) model into its *backend* counterpart.
+    The supported operators and backends can be found at `hummingbird.ml.supported`.
+    """
+    assert model is not None
+    assert torch_installed(), "To use Hummingbird you need to install torch."
+
+    import torch
```
interesaaat: You can remove this.

hummingbird/ml/convert.py
```diff
@@ -221,7 +254,7 @@ def convert(model, backend, test_input=None, device="cpu", extra_config={}):
     extra_config = deepcopy(extra_config)
 
     # Add test input as extra configuration for conversion.
-    if test_input is not None and constants.TEST_INPUT not in extra_config and len(test_input) > 0:
+    if test_input is not None and constants.TEST_INPUT not in extra_config:
```
interesaaat: Did you check the behavior when test_input is `[]`?

hummingbird/ml/convert.py
```diff
@@ -269,4 +340,23 @@ def convert(model, backend, test_input=None, device="cpu", extra_config={}):
     if _is_onnx_model(model):
         return _convert_onnxml(model, backend, test_input, device, extra_config)
 
+    if _is_sparkml_model(model):
+        return _convert_sparkml(model, backend, test_input, device, extra_config)
+
     return _convert_sklearn(model, backend, test_input, device, extra_config)
+
+
+def _is_pandas_dataframe(df):
```
interesaaat: Should we move these 2 into `utils`?

hummingbird/ml/operator_converters/_array_feature_extractor_implementations.py
```diff
@@ -36,3 +36,19 @@ def forward(self, x):
             return x[:, self.min : self.max]
         else:
             return torch.index_select(x, 1, self.column_indices)
+
+
+class VectorAssemblerModel(BaseOperator, torch.nn.Module):
+    """
+    Class implementing ArrayFeatureExtractor in PyTorch
+
+    This is used by Spark0=-ML VectorAssembler
```
interesaaat: typo?

hummingbird/ml/operator_converters/_base_operator.py
```diff
@@ -6,7 +6,7 @@ class BaseOperator(ABC):
     Abstract class defining the basic structure for operator implementations in Hummingbird.
     """
 
-    def __init__(self, regression=False, classification=False, transformer=False, anomaly_detection=False, **kwargs):
+    def __init__(self, regression=False, classification=False, transformer=False, anomaly_detection=False, input_indices=None, append_output=False, **kwargs):
```
interesaaat: Sorry but can you add some documentation for the class? (the signature is a bit bloated and I think it makes sense to have documentation to take track of the different options)

hummingbird/ml/operator_converters/_base_operator.py
```diff
@@ -16,3 +16,31 @@ def __init__(self, regression=False, classification=False, transformer=False, an
         self.classification = classification
         self.transformer = transformer
         self.anomaly_detection = anomaly_detection
+        self.input_indices = input_indices
+        self.append_output = append_output
+
+    # Helper function to select the input for the current operator
+    def select_input_if_needed(self, x):
+        if self.append_output:
+            self.input_x = x
```
interesaaat: what's `input_x`?

interesaaat: If this is used from subclasses, maybe we have it as a property with documentation?

hummingbird/ml/operator_converters/_base_operator.py
```diff
@@ -16,3 +16,31 @@ def __init__(self, regression=False, classification=False, transformer=False, an
         self.classification = classification
         self.transformer = transformer
         self.anomaly_detection = anomaly_detection
+        self.input_indices = input_indices
+        self.append_output = append_output
+
+    # Helper function to select the input for the current operator
```
interesaaat: Period at the end of the comment, please.

hummingbird/ml/operator_converters/sparkml/discretizer.py
```diff
@@ -0,0 +1,57 @@
...
+from .._base_operator import BaseOperator
+from .._discretizer_implementations import Binarizer, KBinsDiscretizer
+
+
+def convert_sparkml_bucketizer(operator, device, extra_config):
+    """
+    Converter for `pyspark.ml.feature.Bucketizer`
+
+    Args:
+        operator: An operator wrapping a `pyspark.ml.feature.QuantileDiscretize` model
```
interesaaat: typo

interesaaat (changes requested):
Two main additional comments:
- we need to add pyspark to the ci\cd pipeline (this should be easy)
- I see that you chose the design where you select the features at runtime. Could't we use the infra we already have in place in the pytorch container? Basically the outer pytorch model will pick the features required for the inner model? I kind of like this because we only select the columns in one single place.

setup.py
```diff
@@ -39,6 +39,7 @@
         "tests": ["flake8", "pytest", "coverage", "pre-commit"],
         "docs": ["pdoc3==0.8.1"],
         "onnx": ["onnxruntime>=1.0.0", "onnxmltools>=1.6.0"],
+        "sparkml": ["pyspark>=3.0.0"],
```
interesaaat: Do you think we can support from spark 2.4.4?

hummingbird/ml/operator_converters/_base_operator.py
```diff
@@ -16,3 +30,31 @@ def __init__(self, regression=False, classification=False, transformer=False, an
         self.classification = classification
         self.transformer = transformer
         self.anomaly_detection = anomaly_detection
+        self.input_indices = input_indices
+        self.append_output = append_output
+
+    # Helper function to select the input for the current operator.
+    def select_input_if_needed(self, x):
+        if self.append_output:
+            self.input_x = x
```
interesaaat: Can we init `self.input_x` in `__init__` with some doc as well? Like this is used by Spark ML to propagate the original inputs.

hummingbird/ml/operator_converters/_base_operator.py
```diff
@@ -16,3 +30,31 @@ def __init__(self, regression=False, classification=False, transformer=False, an
...
+
+        if self.input_indices is not None and (isinstance(x, list) or isinstance(x, tuple)):
+            if len(self.input_indices) > 1:
+                x = tuple([x[i] for i in self.input_indices])
+            else:
+                x = x[self.input_indices[0]]
+        elif isinstance(x, tuple) and len(x) == 1:
```
interesaaat: Can you please add 2 cases:
- one in which `self.input_indices` is `None` and `len(x) > 1`. I think that in this case we should concat all the features. (I probably will need this later on)
- a final else case where we throw a `RuntimeException`. Will this case be ever reached?

tests/test_sparkml_pipeline.py
```diff
@@ -0,0 +1,123 @@
...
+    sc = SparkContext.getOrCreate()
+    sql = SQLContext(sc)
+
+if pandas_installed():
+    import pandas as pd
+
+
+class TestSparkMLPipeline(unittest.TestCase):
+    @unittest.skipIf(not sparkml_installed(), reason="Spark-ML test requires pyspark")
+    @unittest.skipIf(LooseVersion(torch.__version__) < LooseVersion("1.16.0"), reason="Spark-ML test requires torch >= 1.16.0")
```
interesaaat: Does it even exists 1.16? Do you mean 1.6?

interesaaat: I think we also need a skipIf pandas is not installed. I don't think it will ever happen since for this tests we always install everything, but I think it is better to add the check to all tests using it.

interesaaat (commented):
Two questions:
- no way we can do this? <I see that you chose the design where you select the features at runtime. Could't we use the infra we already have in place in the pytorch container? Basically the outer pytorch model will pick the features required for the inner model? I kind of like this because we only select the columns in one single place.
- It looks to me that the tests with pipelines is not triggered because I see from codcov that those lines in _parse are not triggered.

hummingbird/ml/operator_converters/_array_feature_extractor_implementations.py
```diff
@@ -36,3 +36,18 @@ def forward(self, x):
...
+    """
+    Class implementing ArrayFeatureExtractor in PyTorch
+
+    This is used by Spark-ML VectorAssembler
+    """
+
+    def __init__(self):
+        super(VectorAssemblerModel, self).__init__(transformer=True)
+
+    def forward(self, *x):
```
interesaaat: I think that we can point to concat in `pipeline_implementations.py` for this one.

interesaaat (approved):

interesaaat:
Thanks @scnakandala!

PR #319, opened 2020-10-13
Add capability of setting number of threads to the container

### LINKED ISSUE #259: Add support for setting the number of threads directly from the container
Right now users can set the number of thread by manually doing something like (e.g., for torch)
```
torch.set_num_threads(6)
```
It would be good to have this directly into the container so that (1) one does not have to set the number of threads manually, (2) we can provide a consistent, generic API across backends.

interesaaat:
Closes #259.


hummingbird/ml/_container.py
```diff
@@ -6,22 +6,26 @@
 
 """
 All custom model containers are listed here.
+In Hummingbird we use to types of containers:
```
ksaur: s/to/two/g

hummingbird/ml/_container.py
```diff
@@ -40,139 +44,43 @@ def __init__(self, sparkml_model):
...
-                return list(variable_map[output_name] for output_name in self._output_names)
-
-
-class PyTorchTorchscriptSklearnContainer(ABC):
-    """
-    Base container for PyTorch and TorchScript models.
-    The container allows to mirror the Sklearn API.
-    """
+        Base container abstract class allowing to mirror the Sklearn API.
+        Thanks to *SklearnContainer* we can use the `predict`, `predict_proba` etc. API of Sklearn
```
ksaur: Maybe rephrase as
"*SklearnContainer* enables the use of `predict`..."

hummingbird/ml/_container.py
```diff
@@ -40,139 +44,43 @@ def __init__(self, sparkml_model):
...
+        also over the models generated by Hummingbird (irrespective of the selected backend).
 
-    def __init__(self, model, extra_config={}):
-        """
         Args:
-            model: A pytorch or torchscript model
-            extra_config: Some additional configuration parameter
+            model: A model
```
ksaur: maybe "**any** model" or ...something additional here

hummingbird/ml/_container.py
```diff
@@ -40,139 +44,43 @@ def __init__(self, sparkml_model):
...
+            n_threads: How many threads should be used by the containter to run the model. None means use all threads.
+            batch_size: If different than None means split the input into batch_size partitions and score one partition at a time.
```
ksaur: the word "means" is confusing.
"If different than None, split the input into batc..."

hummingbird/ml/_container.py
```diff
@@ -40,139 +44,43 @@ def __init__(self, sparkml_model):
...
+    """
+    Base container for PyTorch and TorchScript models.
+    """
+
+    def __init__(self, model, n_threads=None, batch_size=None, extra_config={}):
+        super(PyTorchTorchscriptSklearnContainer, self).__init__(model, n_threads, batch_size, extra_config)
+
+        assert self._n_threads is not None
+
+        if torch.get_num_interop_threads() != 1:
```
ksaur, line 83: Can you add a comment here on why we must first reset the number of threads to 1?  I know it's the constructor, but at first glance it's confusing.

PR #323, opened 2020-10-14
Add batching to output containers

### LINKED ISSUE #219: Automated batching
From #187, @zhanjiezhu  raised the idea of automated batching.  Using a batch size of 10k rows, for example, can give optimal results.

Right now, the only way to do this is to do batch-processing in a loop with a size a user sets manually.  

We could implement this by adding an option for batching to our [pytorch container](https://github.com/microsoft/hummingbird/blob/2a867101066dda2e007a984a104cad49ee757e56/hummingbird/ml/_container.py#L27), which would facilitate automatically setting up batch processing for users.

(Bonus points if we can figure out how to auto-select the best batchsize based on model/workload/etc)

interesaaat:
Closes #219. This should go after #319.


doc/html/hummingbird/ml/convert.html
```diff
@@ -26,7 +26,7 @@ <h1 class="title">Module <code>hummingbird.ml.convert</code></h1>
 <details class="source">
 <summary>
 <span>Expand source code</span>
-<a href="https://github.com/microsoft/hummingbird/blob/master/hummingbird/ml/convert.py#L0-L271" class="git-link">Browse git</a>
```
ksaur: I think you need to reinstall your precommit hooks; these HTML files should no longer be getting added to PRs.

If you re-run `pre-commit install` it should work; else delete them from `.git/hooks/`

ksaur (commented):
I'm going to try to refresh this diff against the main branch...i think you might need to pull in some of those changes from the prior PR comments?

hummingbird/ml/_topology.py
```diff
@@ -242,6 +248,103 @@ def fix_graph(graph):
         else:
             container = PyTorchSklearnContainerClassification
 
-    hb_model = container(hb_model, extra_config)
+    n_threads = None if constants.N_THREADS not in extra_config else extra_config[constants.N_THREADS]
+    batch_size = None if constants.BATCH_SIZE not in extra_config else extra_config[constants.BATCH_SIZE]
+    hb_model = container(hb_model, n_threads, batch_size, extra_config=extra_config)
```
ksaur: If you set `extra_config[constants.N_THREADS]`, then why is it a separate parameter to `container`?  It looks like you pass `n_threads` to `container` redundantly: both explicitly and as part of the `extra`.
(same comment for batch)

doc/html/hummingbird/ml/convert.html
interesaaat: Yea I think I was still using the version before you removed `pdoc3` from the pipeline.

hummingbird/ml/_topology.py
interesaaat: Good point. I am basically adding the `n_threads` and `batch_size` to the signature of the container (easier to see what is needed for a container to run!?). `extra_config` is needed because some backends need to extra configuration (e.g., tvm).

ksaur:
This is a cool feature! We need to add it to a notebook :)

hummingbird/ml/_container.py
```diff
@@ -67,6 +67,46 @@ def __init__(self, model, n_threads=None, batch_size=None, extra_config={}):
...
+            return self._run_batch_inference(function, *inputs, reshape=reshape)
+
+    def _run_batch_inference(self, function, *inputs, reshape=False):
+        """
+        This function contains the code to run batched inference.
+        """
+
+        if DataFrame is not None and type(inputs[0]) == DataFrame:
+            # Split the dataframe into column ndarrays.
+            inputs = inputs[0]
```
ksaur: codecov claims line 87-91 aren't tested. Can you please either add a case for this, or dig into why codecov is confused?

interesaaat: Ah I don't have a test for DataFrame! Let me add it.

interesaaat: Ok it should be good now. It took a while to get it right, dataframe support is not solid at the moment.

Issue #107, opened 2020-06-04
Automate document generation


KranthiGV:
- We have been manually refreshing documentation (generated using pdoc) till now.
- However, this can be automated using Github actions.

Open questions:
- Do we want to update github pages, too? This may require us to monitor the quality of docstrings closely. This can be a deterrent to a first time contributor.
- What do we do with runtime warnings?
- Is generation of documentation a pre-push or post-push item? If it's post-push, what if we get errors?

cyberShaw:
I am new to open source and would love to work on this issue, can I contribute?

interesaaat:
Welcome @cyberShaw! I assigned this issue to you.

ksaur:
Welcome @cyberShaw! Great questions @KranthiGV, thanks for raising this.

We had originally intended to use Sphinx for documentation, but that setup required a bit more  time than originally planned, so for now we are using pdoc.  (Definitely open to revisiting that decision!)

In terms of order of refreshing content, there's a [nice list](https://sphinxtechnicalwriting.readthedocs.io/en/latest/cd/cdindex.html) on the sphinx page that we could use for our pdoc doc gen flow for now.

`Do we want to update github pages, too?`
My gut reaction is yes; hopefully the docstrings aren't too much of a burden on new contributors. We should definitely write some text about this in CONTRIBUTING.md, as I see that's missing!

`What do we do with runtime warnings...Is generation of documentation a pre-push or post-push item? If it's post-push, what if we get errors?`
I think this is something we might have to experiment with a bit.  I am not familiar with how strict the error tolerance is, I don't want to discourage users.  For example, it's probably ok if the doc is "slightly off-format" vs "totally broken", and I'm not sure where the error threshold falls.

interesaaat:
I did a simple PR (#223) that run pdoc as a pre-commit. In this way, if the code ends up in changing the documentation, the html page with the doc is update automatically. I guess we can go with this solution until we don't have better options. We still need to fix the pipeline for updating the gh-page. That is still manual at the moment.

PR #324, opened 2020-10-14
removing html from commit hooks

ksaur:
We tried to add pdoc into a commit hook (#107), but ended up being more hassle than savings, and we still need to push to the docs branch in the end.


interesaaat:
Question: should we add the string to generate the pdoc documentation [here](https://github.com/microsoft/hummingbird/blob/gh-pages/README.md) so that we don't lose it \ we don't have to dig into the PRs to retrieve how to generate the documentation? 

I am not sure if this is important if we eventually switch to some other format.

ksaur:
> Question: should we add the string to generate the pdoc documentation [here](https://github.com/microsoft/hummingbird/blob/gh-pages/README.md) so that we don't lose it \ we don't have to dig into the PRs to retrieve how to generate the documentation?
> 
> I am not sure if this is important if we eventually switch to some other format.

Ok see #327.

PR #328, opened 2020-10-15
Add benchmark scripts for trees

interesaaat:
Closes #315. This should go after #319 and #323.

PR #330, opened 2020-10-15
Add scripts for the operators bench

### LINKED ISSUE #316: Add benchmarking scripts for operators
The scripts will allow the replication of the result of section 6.1.2 of the paper.

interesaaat:
Closes #316. To merge after #328.


ksaur:
Can you rebase (and remove html) and i'll check it?

hummingbird/ml/_topology.py
```diff
@@ -39,15 +39,23 @@
 else:
     DataFrame = None
 
+if pandas_installed():
```
ksaur, line 42: Something must be wrong with the rebase here.

Also i think there is some other problem with the prior merge... it seems that this [already was added...twice](https://github.com/microsoft/hummingbird/blob/master/hummingbird/ml/_topology.py#L37)

benchmarks/operators/score.py
```diff
@@ -0,0 +1,267 @@
+# Copyright (c) 2019, Microsoft CORPORATION. All rights reserved.
```
ksaur, line 1: Does this file not need BSD/NVid license?  I know you wrote most of this file, just double checking

benchmarks/trees/score.py
```diff
@@ -0,0 +1,257 @@
+# Copyright (c) 2019, Microsoft CORPORATION. All rights reserved.
```
ksaur: (i think this is is only here from not rebase, but same question)

hummingbird/ml/_topology.py
interesaaat: oooh. Let me look into this.

benchmarks/operators/score.py
interesaaat: Yes this file contains only our stuff.

benchmarks/trees/score.py
interesaaat: Same answer :)

PR #334, opened 2020-10-19
Remove requirements on providing input schemas for onnx models

### LINKED ISSUE #305: Generate automatically the schema for ONNX models.
ONNX models contains all the input information therefore it should be possible to automatically generate the schema definition \ inputs starting just from the model. 

If this is true, we can remove `ONNX_INITIAL_TYPES ` from the support input parameters.

interesaaat:
Closes #305.

### DIFF
--- a/hummingbird/ml/convert.py
+++ b/hummingbird/ml/convert.py
@@ -146,64 +146,58 @@ def _convert_onnxml(model, backend, test_input, device, extra_config={}):
     import onnx
 
     # The conversion requires some test input for tracing.
-    # Test inputs can be either provided or generate from the inital types.
-    # Get the initial types if any.
-    initial_types = None
-    if constants.ONNX_INITIAL_TYPES in extra_config:
-        initial_types = extra_config[constants.ONNX_INITIAL_TYPES]
-
+    # Test inputs can be either provided or generate from the input schema of the model.
     # Generate some test input if necessary.
     if test_input is None:
         if backend == onnx.__name__:
             from onnxconverter_common.data_types import FloatTensorType, DoubleTensorType, Int32TensorType, Int64TensorType
 
-            if initial_types is None:
-                # Generate inputs from the ONNX schema.
-                initial_types = []
-                for input in model.graph.input:
-                    name = input.name if hasattr(input, "name") else None
-                    data_type = (
-                        input.type.tensor_type.elem_type
-                        if hasattr(input, "type")
-                        and hasattr(input.type, "tensor_type")
-                        and hasattr(input.type.tensor_type, "elem_type")
-                        else None
+            # Generate inputs from the ONNX schema.
+            initial_types = []
+            for input in model.graph.input:
+                name = input.name if hasattr(input, "name") else None
+                data_type = (
+                    input.type.tensor_type.elem_type
+                    if hasattr(input, "type")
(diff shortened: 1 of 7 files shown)


hummingbird/ml/convert.py
```diff
@@ -73,15 +73,6 @@ def _supported_backend_check_config(model, backend, extra_config):
     if backend is torch.jit.__name__ and constants.TEST_INPUT not in extra_config:
         raise RuntimeError("Backend {} requires test inputs. Please pass some test input to the convert.".format(backend))
 
-    if backend is onnx.__name__:
-        if constants.ONNX_INITIAL_TYPES not in extra_config and constants.TEST_INPUT not in extra_config:
```
interesaaat, line 77: We could remove the support for `ONNX_INITIAL_TYPES`, but the onnx convert depends on it. Should we send a PR there and fix it as we did for the other fix in the API? (I am all for it since we simply our API because we remove one unnecessary extra param).

hummingbird/ml/convert.py
```diff
@@ -164,33 +155,77 @@ def _convert_onnxml(model, backend, test_input, device, extra_config={}):
...
+                    data_type = (
+                        input.type.tensor_type.elem_type
+                        if hasattr(input, "type")
+                        and hasattr(input.type, "tensor_type")
+                        and hasattr(input.type.tensor_type, "elem_type")
+                        else None
+                    )
+                    if name is None:
+                        raise RuntimeError(
+                            "Cannot fetch input name or data_type from the ONNX schema. Please provide some test input."
```
ksaur: If we remove support for `ONNX_INITIAL_TYPES`, this won't be possible right?

hummingbird/ml/convert.py
```diff
@@ -164,33 +155,77 @@ def _convert_onnxml(model, backend, test_input, device, extra_config={}):
...
+                        )
+
+            assert all(
+                map(lambda x: len(x[1].shape) == 2, initial_types)
+            ), "Hummingbird currently support only inputs with len(shape) == 2."
+            first_shape = initial_types[0][1].shape
+            first_type = type(initial_types[0][1])
+            assert all(
+                map(lambda x: x[1].shape == first_shape, initial_types)
+            ), "Hummingbird currently support only inputs with same shape."
```
ksaur: `%s/support/supports/g`

Also, can you add the shapes they provided in the output? onnx error messages are always so obscure to search for. :)

hummingbird/ml/convert.py
ksaur: that would be excellent if we could get that in. I can't see any reason for keeping this other than if the user needs to manually add it for some weird reason...?

hummingbird/ml/convert.py
interesaaat: No no, this will be the default. So we can remove the `if initial_types is None` condition and just generated the types directly from the model

hummingbird/ml/convert.py
interesaaat: You mean checking that the output shape is correct? I am actually by default generating sym dimensions so hopefully we won't have that problem.

hummingbird/ml/convert.py
interesaaat: Yea, I don't know. Ok let me remove it and see what breaks.

PR #336, opened 2020-10-19
move n_features from supported to internal

### LINKED ISSUE #335: Move n_features from supported conf to internal
`N_FEATURES` is not in `supported.py` but I think it should not be surfaced to users but just used internally (it is much easier to pass a test dataset rather than having specific configurations for the each algorithm).

interesaaat:
Closes #335.


ksaur (approved):
Should I merge this one now, or wait?

interesaaat:
This one is just a small fix. So it should be fine to merge now, hopefully it does not create too many problems!

PR #339, opened 2020-10-21
add support for onnx models with multiple inputs

interesaaat:


hummingbird/ml/convert.py
```diff
@@ -193,27 +193,33 @@ def _convert_onnxml(model, backend, test_input, device, extra_config={}):
...
+                    raise RuntimeError(
+                        "Type {} not supported. Please fill an issue on https://github.com/microsoft/hummingbird/.".format(
+                            type(it[1])
+                        )
                     )
-                )
+            if extra_config[constants.N_INPUTS] == 1:
+                test_input = test_input[0]
+            else:
+                test_input = tuple(test_input)
```
ksaur, line 222: Do we have a test case for this line 222?    It doesn't look like something likely to fail,  but since it's new i thought I'd check

interesaaat: I think I fixed it and I also added an additional test for the exception that was not covered.

interesaaat: Mmm somehow it is not hit. I am sure that the exception is raise in the new test, so no idea what is going on with the codcov.

interesaaat: Ok I added the test.

PR #343, opened 2020-10-23
Using Sphinx for docs/web

### LINKED ISSUE #325: Doc automation
In #107, we looked at automating our doc generation as part of our build, but removed it in #324 because it still requires pushing the html to a separate branch, and generally caused PRs to be larger than necessary with all of the HTML code.

Ideally we need a system that auto-generates and pushes the docs (maybe as part of our release process rather than every commit).

And we also need to revisit onboarding to sphinx now that we are growing.

ksaur:
Closes #325


ksaur:
@interesaaat - I opened #344 about the weirdness with `torch` that caused this morning's pipeline battle.  

I *think* that if we merge this to `master` that the final step of the pipeline will fire (`name: Deploy to GitHub pages`).

.github/workflows/pythonapp.yml
```diff
@@ -84,3 +84,14 @@ jobs:
       with:
         file: ./coverage.xml
         flags: unittests
+    - name: Generate Documentation   # for some awful reason, this only works with older torch????
+      if: ${{ matrix.python-version == '3.7' && matrix.os == 'ubuntu-latest' }}
```
interesaaat, line 88: should this be 3.5?

.github/workflows/pythonapp.yml
```diff
@@ -84,3 +84,14 @@ jobs:
...
+      run: |
+        make sphinx-site -C website/
+    - name: Deploy to GitHub pages
+      if: ${{ github.ref == 'refs/heads/master' && matrix.python-version == '3.7' && matrix.os == 'ubuntu-latest' }}
```
interesaaat, line 92: This one too.

.github/workflows/pythonapp.yml
ksaur: no, I got it working across all versions...finally.  :)   And since Python3.5 will be deprecated in January, we don't want to use that.

Per our convo this morning, we'll just have one of the targets do the coverage, so i figured Ubuntu  w/3.7 could be the machine that does coverage. But maybe that makes this  target too long.  I'm not sure what other "matrix" ci/cd systems do for things that should  only be run once

interesaaat (approved):

PR #349, opened 2020-10-28
pin versions of install requires deps

interesaaat:
This should also test that pytorch 1.7 now works

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install pytorch for mac
       if: ${{ matrix.os == 'macos-latest' }}
       run: |
-        pip install torch
+        pip install torch==1.7.0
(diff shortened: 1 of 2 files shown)


.github/workflows/pythonapp.yml
```diff
@@ -35,10 +35,10 @@ jobs:
       if: ${{ matrix.python-version == '3.5' && matrix.os != 'macos-latest' }}
       run: |
         pip install torch==1.5.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
-    - name: Install pytorch 1.6.0+cpu if python > 3.5 (not mac)
+    - name: Install pytorch 1.7.0+cpu if python > 3.5 (not mac)
```
ksaur, line 42: do we need to also pin line 33 to a specific version?

interesaaat: yes! Good catch!

PR #353, opened 2020-10-28
Binarizer for ONNX-ML

ksaur:
To be merged AFTER 0.1.0 release.

### DIFF
--- a/tests/test_onnxml_binarizer_converter.py
+++ b/tests/test_onnxml_binarizer_converter.py
@@ -85,7 +85,6 @@ class TestONNXBinarizer(unittest.TestCase):
 
         # generate test input
         onnx_ml_model = convert_sklearn(model, initial_types=[("float_input", FloatTensorType_onnx(X.shape))])
-        print(onnx_ml_model.graph.node[0].attribute[0])
         onnx_ml_model.graph.node[0].attribute[0].name = "".encode()
 
         self.assertRaises(RuntimeError, convert, onnx_ml_model, "onnx", X)
--- a/tests/test_onnxml_binarizer_converter.py
+++ b/tests/test_onnxml_binarizer_converter.py
@@ -47,26 +47,48 @@ class TestONNXBinarizer(unittest.TestCase):
     @unittest.skipIf(
         not (onnx_ml_tools_installed() and onnx_runtime_installed()), reason="ONNXML test requires ONNX, ORT and ONNXMLTOOLS"
     )
+    # Check 0.0 threshold
     def test_binarizer_converter_0thresh(self, rtol=1e-06, atol=1e-06):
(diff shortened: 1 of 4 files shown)


tests/test_onnxml_binarizer_converter.py
```diff
@@ -0,0 +1,95 @@
...
+    # if the model is corrupt, we should get a RuntimeError
+    def test_onnx_binarizer_converter_raises_rt(self):
+        warnings.filterwarnings("ignore")
+        X = np.array([[1, 2, 3], [4, 3, 0], [0, 1, 4], [0, 5, 6]], dtype=np.float32)
+        model = Binarizer(threshold=0)
+        model.fit(X)
+
+        # generate test input
+        onnx_ml_model = convert_sklearn(model, initial_types=[("float_input", FloatTensorType_onnx(X.shape))])
+        print(onnx_ml_model.graph.node[0].attribute[0])
```
interesaaat: Is this print required?

tests/test_onnxml_binarizer_converter.py
```diff
@@ -0,0 +1,95 @@
...
+        onnx_ml_pred, onnx_pred = self._test_binarizer_converter(-2.0)
+
+        # Check that predicted values match
+        np.testing.assert_allclose(onnx_ml_pred, onnx_pred, rtol=rtol, atol=atol)
+
+    @unittest.skipIf(
+        not (onnx_ml_tools_installed() and onnx_runtime_installed()), reason="ONNXML test requires ONNX, ORT and ONNXMLTOOLS"
+    )
+    # if the model is corrupt, we should get a RuntimeError
+    def test_onnx_binarizer_converter_raises_rt(self):
```
interesaaat, line 80: Should we have this test in some other base test? I think it is independent from binarizer, no?

tests/test_onnxml_binarizer_converter.py
ksaur: good catch, will remove

tests/test_onnxml_binarizer_converter.py
ksaur: This hits line 36 of `binarizer.py`.  It's specific to `Binarizer` in that it makes sure the threshold is set on lines 31-33.  (Mostly, I wanted 100% coverage :-D )

Q- Is this exception case realistic?  Alternatively, I could remove lines 35-36 of`binarizer.py` since I am not sure they can actually ever be reached unless a user specifically corrupts the object like I did (line 88) for the  test. :)

interesaaat: I don't know! I would say to leave it, better be conservative.

interesaaat (approved):

PR #361, opened 2020-11-09
Adding support for Spark ML SQLTransformer

### LINKED ISSUE #318: Add support for Spark ML SQLTransformer
Explore the possibility of adding support for Spark ML SQLTransfomer (https://spark.apache.org/docs/latest/ml-features.html#sqltransformer) in HB.

scnakandala:
Closes #318


hummingbird/ml/operator_converters/sparkml/sql_transformer.py
```diff
@@ -0,0 +1,116 @@
...
+        self.project_list = project_list
+
+    def forward(self, *x):
+        inputs_dict = {name : tensor for name, tensor in zip(self.input_names, x)}
+        # The first node of the project_node_def is the name definition node. Hence ignored.
+        return [self.calculate_output_feature(project_node_def_stack[1:], inputs_dict)[0] for project_node_def_stack in self.project_list]
+
+    def calculate_output_feature(self, project_node_def_stack, inputs_dict):
+        if project_node_def_stack[0]['class'] == 'org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute':
+            return inputs_dict[project_node_def_stack[0]['nameParts'][1:-1].strip()], project_node_def_stack[1:]
```
interesaaat: Can we add some documentation on the different cases? Like what is this doing?

hummingbird/ml/operator_converters/sparkml/sql_transformer.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Returns:
+        A PyTorch model
+    """
+    from pyspark.sql import SparkSession
+
+    spark = SparkSession.builder.getOrCreate()
+    parser = spark._jsparkSession.sessionState().sqlParser()
+    statement = operator.raw_operator.getStatement()
+    plan = parser.parsePlan(statement)
```
interesaaat: This is "just" a parsed plan, right? Meaning the query it is not actually optimized. In general I think it is ok, but bonus point if we can somehow get the optimized logical plan. I think this will also solve the problems with `unresolvedAttribute`, `stars`, etc.

hummingbird/ml/operator_converters/sparkml/sql_transformer.py
```diff
@@ -0,0 +1,116 @@
...
+    plan_json = json.loads(plan.toJSON())
+
+    project_node = [node for node in plan_json if node['class'] == 'org.apache.spark.sql.catalyst.plans.logical.Project']
+    if len(project_node) == 0:
+        raise RuntimeError('Couldn\'t find the projectList from the SQLTransformer statement: {}.'.format(statement))
+
+    # Ignores select *
+    project_list = [proj for proj in project_node[0]['projectList'] if proj[0]['class'] != 'org.apache.spark.sql.catalyst.analysis.UnresolvedStar']
+    input_names = [input.raw_name for input in operator.inputs]
+    return SQLTransformerModel(input_names, project_list, device)
```
interesaaat: Ok so here you are generating a uber operator for the whole `SQLTransformer`. Another possibility is to use `SQLTransformer` as a specific case in `_parse` and add each single operator one by one into the topology. Have you thought about this other option? Do you have an idea on the possible pros and cons between these 2 possibilities?

tests/test_sparkml_sql_transformer.py
```diff
@@ -0,0 +1,54 @@
...
+        pd_df = pd.DataFrame(data=np.c_[iris['data'], iris['target']], columns=features + ['target'])[["sepal_length", "sepal_width", "petal_length", "petal_width"]]
+        df = sql.createDataFrame(pd_df)
+
+        model = SQLTransformer(statement="SELECT *, sepal_length*(sepal_length/sepal_width) as new_feature1,"
+                                         " petal_length*(petal_length/petal_width) as new_feature2,"
+                                         " petal_length + 10.0*(sepal_width - petal_width) as new_feature3 from __THIS__")
+
+        output_col_names = ['new_feature1', 'new_feature2', 'new_feature3']
+        test_df = df
+        torch_model = convert(model, "torch", test_df)
```
interesaaat, line 82: One thing that I noticed in other Spark tests is that we use `test_df` at conversion time (i.e., a Spark `DataFrame`) and a regular pandas `DataFrame` at inference. We should make sure that also inference works over Spark DataFrames (not necessarily in this PR).

hummingbird/ml/operator_converters/sparkml/sql_transformer.py
scnakandala: Right now the parsing logic is mostly focussed on topology of the DAG and the corresponding inputs and outputs. The semantics of the operator is handled at the actual operator definition. We can treat each project column as a different operator and handle it at the _parse level. But that will require elevating some of the operator level logic to the parsing logic. WDYT?

tests/test_sparkml_sql_transformer.py
scnakandala: Agree! However, I think that will require an integration of HB with Spark to distribute the inference using Spark on Spark DataFrames. Also, I think we should support the Pandas data frame at conversion time. Some people may only have the serialized Spark ML pipeline and not the original data frame.

interesaaat: Yes I agree. Actually that would be simpler, right? We can just assume that the input of the converter \ transformation is a Pandas DataFrame, and call it done. Perhaps we can add an exception in the case it's a Spark DataFrame asking to convert it a Pandas DataFrame.

hummingbird/ml/operator_converters/sparkml/sql_transformer.py
interesaaat: I see your point. If we add the operator in _parse we are basically exploding it and merging the Spark SQL plan DAG into our topology DAG. This is similar to how we explode the `ColumnTransformer` or `Pipeline` object in sklearn models. So basically we need a parsing function specific for Spark SQL plans that (1) opens the plan, (2) traverse the plan, (3) transform each operator in the plan to a node in the topology object, and (4) wires the inputs\outpus.

scnakandala:
I think it boils down to the implementation. For example, using TVM we should be able to implement reproducible scatter reductions using reduction factoring and cross-thread reductions (https://tvm.apache.org/docs/tutorials/language/reduction.html).

Without direct scatter reductions we need to materialize a large intermediate tensor and do the reduction. In the group by aggregates it will be |G| * N tensor.

interesaaat:
> I think it boils down to the implementation. For example, using TVM we should be able to implement reproducible scatter reductions using reduction factoring and cross-thread reductions (https://tvm.apache.org/docs/tutorials/language/reduction.html).
> 
> Without direct scatter reductions we need to materialize a large intermediate tensor and do the reduction. In the group by aggregates it will be |G| * N tensor.

Thanks to the pointer to TVM reduction, I was not aware of it. This is a bit more low level than the regular HB approach but we can come up with something if this gives us the best performance.

PR #366, opened 2020-11-17
[Benchmark] Support XGBoost GPU prediction and remove large CPU overhead in benchmark timing

### LINKED ISSUE #363: [Benchmark] Support XGBoost GPU prediction and remove large CPU overhead in benchmark timing
XGB supports GPU prediction, so we should be able to compare XGB GPU perf and HB + GPU on Torch, TVM.

Also, I found that there is a large CPU overhead in how prediction results are stored in the `predictions` array. 
https://github.com/microsoft/hummingbird/blob/main/hummingbird/ml/_container.py#L103-L115

Before the benchmarking script was merged to the main branch, `predictions` array was preallocated and results were written directly to it, like this
```
self.predictions = np.empty([total_size, self.params["n_classes"]], dtype="f4")for i in range(0, iterations):
for i in range(iterations)
    start = i * batch_size
    end = min(start + batch_size, total_size)
    self.predictions[start:end, :] = self.model.predict_proba(predict_data[start:end, :])
```  

This seemingly minor difference results in a serious perf regression for TVM results, compared to numbers I got last month.

I have a WIP branch for this, will send a PR next week.
https://github.com/masahi/hummingbird/tree/xgb-fix

masahi:
Closes https://github.com/microsoft/hummingbird/issues/363

### DIFF
--- a/hummingbird/ml/_container.py
+++ b/hummingbird/ml/_container.py
@@ -48,7 +48,7 @@ class CommonSparkMLModelContainer(CommonSklearnModelContainer):
 # Output containers.
 # Abstract containers enabling the Sklearn API.
 class SklearnContainer(ABC):
-    def __init__(self, model, n_threads=None, batch_size=None, extra_config={}):
+    def __init__(self, model, num_output_columns, n_threads=None, batch_size=None, extra_config={}):
         """
         Base container abstract class allowing to mirror the Sklearn API.
         *SklearnContainer* enables the use of `predict`, `predict_proba` etc. API of Sklearn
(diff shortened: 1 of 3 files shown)


hummingbird/ml/_container.py
```diff
@@ -112,11 +118,17 @@ def _run_batch_inference(self, function, *inputs, reshape=False):
             # Tell function that we are in the last iteration and do proper actions in case
             # (e.g., for TVM we may want to use the raminder model).
             self._last_iteration = i == iterations - 1
-            predictions.extend(function(*batch).ravel())
+            out = function(*batch)
+            out = np.expand_dims(out, 1) if len(out.shape) == 1 else out
+
+            if predictions is None:
```
masahi, line 124: I had to add this check later for the case when `TEST_INPUT` is not provided (`torch` backend), but then I realized that if I have this I don't need `num_output_columns` or `output_index` changes...

masahi:
XGB GPU is added in https://github.com/microsoft/hummingbird/pull/369

I'll work on a new PR to address the CPU overhead issue, following https://github.com/microsoft/hummingbird/issues/368

PR #369, opened 2020-11-21
[Benchmark] Add batch_benchmark option to do benchmark on a single batch (not on a whole data)

masahi:
Also removed batch by batch prediction code in rf/xgb/lgm, since they can do prediction on the whole data in one go and that's how people use them.

If you want to benchmark scaling behavior wrt a specific batch size, you can use the new `batch_benchmark` option instead. I don't think it justifies to do batch by batch prediction just for the sake of doing prediction on the whole data. 

I think measuring the prediction time on the whole data vs benchmarking the scaling behavior wrt batch size are separate issues and we shouldn't conflate them.

@interesaaat

### DIFF
--- a/benchmarks/trees/run.py
+++ b/benchmarks/trees/run.py
@@ -215,7 +215,7 @@ def benchmark(args, dataset_folder, model_folder, dataset):
         results[op] = {}
         model_name = op + "-" + str(args.ntrees) + "-" + str(args.cpus)
         model_full_name = os.path.join(model_folder, model_name + ".pkl")
-        trainer = train.TrainEnsembleAlgorithm.create(op)
+        trainer = train.TrainEnsembleAlgorithm.create(op, data.learning_task)
         test_data = get_data(data.X_test)
 
         with trainer:
--- a/benchmarks/trees/run.py
+++ b/benchmarks/trees/run.py
@@ -187,7 +187,7 @@ def parse_args():
         action="store_true",
         help=("Whether to do a single batch benchmark with specified batch_size and niters (not on the whole data)"),
     )
-    parser.add_argument("-maxdepth", default=None, type=int, help=("Maxmimum number of levels in the trees"))
+    parser.add_argument("-max_depth", default=8, type=int, help=("Maxmimum number of levels in the trees"))
     parser.add_argument(
         "-validate", default=False, action="store_true", help="Validate prediction output and fails accordingly."
     )
@@ -195,7 +195,7 @@ def parse_args():
     args = parser.parse_args()
     # Default value for output json file.
     if not args.output:
-        args.output = "result-{}-{}-{}.json".format("gpu" if args.gpu else args.cpus, args.ntrees, args.batch_size)
+        args.output = "result-{}-{}-{}-{}.json".format("gpu" if args.gpu else args.cpus, args.ntrees, args.max_depth, args.batch_size)
     return args
 
 
@@ -222,7 +222,7 @@ def benchmark(args, dataset_folder, model_folder, dataset):
     for op in operators.split(","):
         print("Running '%s' ..." % op)
         results[op] = {}
-        model_name = op + "-" + str(args.ntrees) + "-" + str(args.cpus)
+        model_name = op + "-" + str(args.ntrees) + "-" + str(args.max_depth) + "-" + str(args.cpus)
         model_full_name = os.path.join(model_folder, model_name + ".pkl")
         trainer = train.TrainEnsembleAlgorithm.create(op, data.learning_task)
         if args.batch_benchmark:
--- a/benchmarks/trees/run.py
+++ b/benchmarks/trees/run.py
@@ -193,6 +193,14 @@ def parse_args():
     return args
 
 
+def get_data(data, size=-1):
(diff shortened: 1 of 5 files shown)


benchmarks/trees/train.py
```diff
@@ -46,58 +42,37 @@
...
-                self.predictions = np.empty([total_size], dtype="f4")
-                predict_fn = model.predict
-
-            for i in range(0, iterations):
-                start = i * batch_size
-                end = min(start + batch_size, total_size)
-                batch = TrainEnsembleAlgorithm.get_data(predict_data, start, end)
-                self.predictions[start:end] = predict_fn(batch)
-
+            self.predictions = predict_fn(predict_data)
```
masahi: This change may be controversial, but I strongly believe we shouldn't add batching to rf/xgb/lgm, since they can do prediction in one go. Batching is only needed for HB + some backends that require fixed input shape (TVM), so it is only HB's minor technical concern and shouldn't affect rf/xgb/lgm.

Adding batching also to rf/xgb/lgm for the sake of "apple to apple comparison" makes things considerably slower for them, so I don't think this is a fair benchmark. Moreover, by allowing TVM to compile for a fixed shape, which enables many static optimization, I argue that this benchmark is already not an "apple to apple comparsion".

We should compare the best rf/xgb/lgm perf vs the best HB perf, since that's what people care about. If we report rf/xgb/lgm numbers with batch by batch prediction, that wouldn't reflect the numbers people would see in practice.  

@interesaaat @ksaur

interesaaat: I don't agree with this. I think we should have both, and benchmark on both. We want to support (and benchmark) batching because we want (1) out of memory computations (which are not supported by sklearn/xgb/lgm); (2) an easy way to benchmark online (single record) inference.

Instead of removing this, we can just use the `batch_benchmark` flag: if enabled we do batching (as the code is set now), if not, we just do inference over the whole dataset.

benchmarks/trees/run.py
```diff
@@ -181,6 +181,12 @@ def parse_args():
         ),
     )
     parser.add_argument("-niters", default=5, type=int, help=("Number of iterations for each experiment"))
+    parser.add_argument(
```
interesaaat, line 184: Is this actually needed? I think it's a bit redundant since we can we use something like `batch_size <0` to trigger this behavior. I don't have really strong opinions on this, if you think it is easier for having "no batching experiments".

benchmarks/trees/train.py
```diff
@@ -138,6 +113,13 @@ def configure(self, data, args):
     def fit(self, data, args):
         params = self.configure(data, args)
 
+        tree_method = params["tree_method"]
+        predictor = "cpu_predictor"
+
+        if args.gpu:
+            tree_method = "gpu_hist"
+            predictor = "gpu_predictor"
```
interesaaat, line 144: Thanks for adding this!

interesaaat (changes requested):

benchmarks/trees/run.py
masahi: Yes this is needed. My intention is, with `batch_benchmark=true`, we want to benchmark on the batch size specified in `batch_size` argument (10000 by default), so `batch_size < 0` doesn't make sense unless you are saying we can use the whole dataset as one batch. I think it is better to be able to do benchmark on different batch size (1, 100, 10000 etc).

benchmarks/trees/train.py
masahi: ok I'll remove this change since this is not essential for this PR and it needs further discussion. I can see that for out of memory prediction benchmark use case, having batching in xgb makes sense (assuming `xgb_model.predict(huge_data)` in one go doesn't work). Note that  the second point you mentioned (single record bench) is better handled by the new `batch_benchmark` option I added in this PR with batch size = 1.

But I'd say that is rather special use case - this is not something we should benchmark **by default**. You are essentially saying, by default when you run `python run.py`, which means batch size = 10000 and prediction is done batch by batch on the whole dataset, you are benchmarking for the hypothetical, out of core use case where we decide to run batch by batch inference with batch size 10000. Causal users might run `python run.py` (which means out of core benchmark) and get the impression that HB is 5x faster, say, when it is not.

So for the benchmark that runs by default, I think it is better to do a single batch benchmark using the new option I added in the PR, or benchmark on the whole dataset as one batch. Either way, batching in rf/xgb won't take effect so we get the best result they can achieve. On the HB + TVM side, we can do either batch / no batch, depending on the `batch_size` argument. I believe this is the most fair benchmark and these are the numbers we should report by default.

masahi: Restored batching for xgb

masahi:
@interesaaat I think this is ready for merge. Can you take another look? I've also added fixes to a few issues I found during benchmarking.

Summary of the changes:
* Add `batch_benchmark` option
* Enable GPU support in xgb
* Bug fix where `max_depth` argument is ignored
* Create ORT session once in `ONNXML` scorer

interesaaat:
Yes, thanks! Will take a look tomorrow morning.

interesaaat (approved):

interesaaat:
Thanks @masahi for the contribution!

Issue #160, opened 2020-06-19
Cleanup SKL LabelEncoder Converter Code


ksaur:
We have the original draft of the SKL LabelEncoder code uploaded in [this branch](https://github.com/microsoft/hummingbird/tree/cleanup/label-encoder). All credit here goes to @scnakandala, the original author and brains behind this.

It contains an un-edited implementation of LabelEncoder [here](https://github.com/microsoft/hummingbird/blob/cleanup/label-encoder/hummingbird/ml/operator_converters/skl_label_encoder.py).

There is a test file [here](https://github.com/microsoft/hummingbird/blob/cleanup/label-encoder/tests/test_sklearn_label_encoder_converter.py) that needs to be cleaned up and passing (and also tests added for errors).

**Notes:**

* We had to treat strings a bit differently because of the way they are represented in tensors
  * We should add more expansive tests
* In the numeric label encoder, there is a tricky spot [here](https://github.com/microsoft/hummingbird/blob/cleanup/label-encoder/hummingbird/ml/operator_converters/skl_label_encoder.py#L46) related to running on GPU vs CPU.  
  * There may be some bugfixes/improvements that are available now. Our CI/CD pipeline is not currently GPU-enabled, but it would be helpful to be able to run this on a GPU-enabled machine when developing.

amueller:
Btw, many people that are using ``LabelEncoder`` should actually be using ``OneHotEncoder``. ``LabelEncoder`` was meant for labels i.e. targets, not features, and only supports a single column, while ``OneHotEncoder`` supports multiple columns.

interesaaat:
Good to see you here @amueller! :) Yes we also have `OneHotEncoder`, it is work in progress in PR #147.

ksaur:
closed via #374

PR #374, opened 2020-12-03
onnxml and SKL Label Encoder

ksaur:
This adds LabelEncoder for SKL (strings and ints) and ONNX (ints).

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, macos-latest, windows-2019]
+        os: [ubuntu-18.04, macos-10.14, windows-2019]
         python-version: [3.5, 3.6, 3.7, 3.8]
 
     steps:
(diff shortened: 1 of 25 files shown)


ksaur:
closes #160

ksaur:
Ok i think we have a good hybrid to be prepared for torch==1.8.  PTAL @interesaaat !

.github/workflows/pythonapp.yml
```diff
@@ -47,9 +47,9 @@ jobs:
     - name: Install pytorch 1.5.1+cpu if python 3.5 (not mac)
       if: ${{ matrix.python-version == '3.5' && matrix.os != 'macos-10.14' }}
       run: pip install torch==1.5.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
-    - name: Install pytorch 1.7.0+cpu if python > 3.5 (not mac)
+    - name: Install pytorch nightly if python > 3.5 (not mac)
       if:  ${{ matrix.python-version != '3.5' && matrix.os != 'macos-10.14' }}
-      run: pip install torch==1.7.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+      run: pip install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
```
interesaaat: I am not sure we want to run our cicd on their nightly built. Since it's not stable we may get intro troubles.

tests/test_onnxml_label_encoder_converter.py
```diff
@@ -0,0 +1,149 @@
...
+        ]
+        model.fit(data)
+
+        # max word length is the smallest number which is divisible by 4 and larger than or equal to the length of any word
+        max_word_length = 12
+        view_args = [-1, max_word_length // 4]
+        str_dtype = "|S" + str(max_word_length)
+        np_arr = np.array(data, dtype=str_dtype).view(np.int32)
+
+        # pytorch_input = torch.from_numpy(np_arr).view(*view_args)
```
interesaaat, line 79: This can be removed

interesaaat (approved):
Just have 2 small comments. Do you want to fix them, or I can.

.github/workflows/pythonapp.yml
ksaur: ok, since we need 1.8 but it's not in pypi yet, do you suggest we pin it to a [specific version](https://download.pytorch.org/whl/nightly/cu110/torch_nightly.html) of 1.8? If so, which?

interesaaat: We need 1.8 only for exporting to onnx, pytorch should still work (TVM, I don't know). So I think it is safer to leave 1.7 in the pipeline and move to 1.8 once it's stable. When the onnx export is used, we can either leave as it is (so it will say `nonzero` is not currently supported), or we can add a specific exception, like label encoder not supported until pytorch 1.8.

ksaur: k sounds good

Issue #368, opened 2020-11-20
Batch inference API for backends that require fixed batch size (TVM)


masahi:
TVM backend requires a fixed batch size specified at convert time. Currently, hummingbird uses [_run_batch_inference](https://github.com/microsoft/hummingbird/blob/main/hummingbird/ml/_container.py#L89) function to run prediction on test data whose batch size is bigger than the batch size TVM is compiled for.

`hb_model.predict(test_data)` is supposed to work the same way as sklearn does, and that is taken care of by `_run_batch_inference` above. But this has two problems for TVM backend, in terms of CPU overhead spent outside of TVM:

* It requires allocating a large array to hold each fixed-size batch prediction (related PR https://github.comhttps://github.com/microsoft/hummingbird/pull/366/microsoft/hummingbird/pull/366)
* And copying each batch prediction to the result array

For the first item, during the work of https://github.com/microsoft/hummingbird/pull/366 I found that pre-allocating a prediction array inside `_run_batch_inference` is not straightforward, since it is used by all prediction tasks. For classification and regression it is no problem, but for transform it doesn't seem to work. So I want to close that PR and instead deal with the CPU overhead problem in a different way.

For performance reason, the following change is desired:
* `tvm_model.predict(test_data)` works only for the same batch size TVM is compiled for. This means `tvm_model.predict(test_data)` is just a thin wrapper around TVM's `model.run()`. 
* For test data whose size is larger than the fixed batch size, introduce another API to return a list of prediction (as a list of ndarray), rather than a one big array. 

One possibility is to add `predict_batch(test_data)` API whose signature is ndarray -> [ndarray].
`predict(test_data)` works the same way as in current implementation, but a restriction needs to be added for TVM backend: the test data to be passed in needs to have the same batch size specified at compile time.

Let me know your thoughts @interesaaat @ksaur

interesaaat:
I like this idea. Another problem with the current setup is that if the final prediction size is bigger than memory, we will fail. With this proposal we can solve also this issue as well as the performance problem. 

One minor point that I need to think about is the `predict_batch` API. I am not 100% sure that having a different signature will help here. I am a bit worried that we will add a technical detail (i.e., batching) to the main predict API will create confusion.
[...]

PR #376, opened 2020-12-04
Refactor TVM compilation and prediction code

masahi:
This is just a simple refactoring of TVM related code in preparation for https://github.com/microsoft/hummingbird/issues/368

@interesaaat


interesaaat:
This looks good! Thanks for the refactoring. Looking forward for the next step.

PR #377, opened 2020-12-04
Introduce BatchContainer for batch by batch prediction use case

masahi:
The proposed implementation for https://github.com/microsoft/hummingbird/issues/368

All batch by batch prediction logic, including `remainder_model` stuff, is encapsulated in the new `BatchContainer` class.

The new function `convert_batch` is supposed to be used to get `BatchContainer`:
`convert_batch(model, backend, test_input, remainder_size=0, device="cpu", extra_config={})`

Note that `test_input` is a required argument. `test_input.shape[0]` is interpreted as `batch_size`, and `test_input` is used for tracing.

The key idea is, I create a new container for `remainder_model`. So individual containers only have one underlying compiled TVM model, no need to have both "main" model and "remainder" in one container. See below:
  
https://github.com/masahi/hummingbird/blob/2890709b5f6828d91a2f7f53e4dab0727d5c6fca/hummingbird/ml/_topology.py#L379-L388

And `BatchContainer` takes "main" container, which is used for batch by batch prediction except the last iteration, and optionally `remainder_model_container` when `remainder_size = 0` that is used for the last iteration.
```
class BatchContainer:
    def __init__(self, base_container, remainder_model_container=None):
```

All `predict` related methods of `BatchContainer` takes an optional `concatenate_outputs` argument, which is true by default. If True, we concatenate the list of predictions into one array, so the `BatchContainer` acts just like sklearn. Users who care about the highest performance can pass `concatenate_outputs=False` to bypass the potentially expensive final concatenation and get a list of predictions. 

```
class BatchContainer:
    ...
    def predict(self, *inputs, concatenate_outputs=True):
          ...
```

The changes to the existing `convert(...)` API:
* `test_input` is directly used as tracing input. `test_input.shape[0]` is interpreted as `batch_size`.
* BATCH_SIZE argument is ignored.
* For TVM backend, the container returned by `convert(...)` can only predict on the data with the same shape as `test_input`. An assertion is added to make sure of this.

`convert_batch` API usage is demonstrated in `test_extra_conf.py`. All tests in `test_extra_conf.py` pass. I'll fix other tests we agree on the implementation.

@interesaaat


hummingbird/ml/_container.py
```diff
@@ -512,21 +581,19 @@ def __init__(self, model, n_threads=None, batch_size=None, extra_config={}):
...
         self._to_tvm_array = lambda x: tvm.nd.array(x, self._ctx)
 
         os.environ["TVM_NUM_THREADS"] = str(self._n_threads)
 
     def _to_tvm_tensor(self, *inputs):
-        return {self._input_names[i]: self._to_tvm_array(inputs[i]) for i in range(len(inputs))}
+        tvm_tensors = {}
+        msg = "The number of input rows %d is different from the batch size %d the TVM model is compiled for."
+        for i, inp in enumerate(inputs):
+            assert inp.shape[0] == self._batch_size, msg % (inp.shape[0], self._batch_size)
```
interesaaat: Nit: throughout the codebase we always use `.format` instead of `%`. I would like to be consistent. What is the convention here in python projects?

hummingbird/ml/convert.py
```diff
@@ -400,3 +368,48 @@ def convert(model, backend, test_input=None, device="cpu", extra_config={}):
...
+        >>> pytorch_model = convert(sklearn_model,`torch`)
+
+    Returns:
+        A model implemented in *backend*, which is equivalent to the input model
+    """
+    assert constants.REMAINDER_SIZE not in extra_config
+    return _convert_common(model, backend, test_input, device, extra_config)
+
+
+def convert_batch(model, backend, test_input, remainder_size=0, device="cpu", extra_config={}):
```
interesaaat, line 416: Please make sure the documentation will be clear on explaining what `remainder_size`means.

hummingbird/ml/convert.py
```diff
@@ -95,7 +95,7 @@ def _convert_sklearn(model, backend, test_input, device, extra_config={}):
     topology = parse_sklearn_api_model(model, extra_config)
 
     # Convert the Topology object into a PyTorch model.
-    hb_model = topology_converter(topology, backend, device, extra_config=extra_config)
+    hb_model = topology_converter(topology, backend, test_input, device, extra_config=extra_config)
```
interesaaat, line 98: We passed `test_input` into `extra_config` because it is not always used and because some op will require it anyway among the `extra_config`

hummingbird/ml/_topology.py
```diff
@@ -372,10 +373,21 @@ def fix_graph(graph):
...
+
+    if remainder_model:
+        aux_container = container(remainder_model, n_threads, remainder_size, extra_config=extra_config)
+        return BatchContainer(hb_container, aux_container)
+    elif remainder_size is not None and remainder_size > 0:
+        # torch backend case
+        aux_container = container(hb_model, n_threads, remainder_size, extra_config=extra_config)
+        return BatchContainer(hb_container, aux_container)
+    elif remainder_size is not None:
+        assert remainder_size == 0, "remainder_size is non zero but no remainder_model has been created"
```
interesaaat, line 387: What is this case covering? When `remainder_size` is equal to `batch_size`?

interesaaat:
Looks ok so far. I only had minor comments. I didn't look too much into the details, I will take a better look once complete. I saw that you are also fixing some other stuff along the way, thanks!

hummingbird/ml/_container.py
masahi: Replaced with format. Yes format is more common I think.

hummingbird/ml/_topology.py
masahi: No, this is for the case when `remainder_size` is not `None` and remainder_size == 0. I make sure that `remainder_size` is not `None` only if `convert_batch` is called, see

https://github.com/microsoft/hummingbird/blob/2eda7538106888d56b6a3b1d1a70c0d33fa15354/hummingbird/ml/convert.py#L406
https://github.com/microsoft/hummingbird/blob/2eda7538106888d56b6a3b1d1a70c0d33fa15354/hummingbird/ml/convert.py#L414

In the case `remainder_size` is 0, no `remainder_model` is compiled. But since a user has asked for `BatchContainer` using `convert_batch`, we return `BatchContainer` but without the container for `remainder_model`.

masahi:
@interesaaat All tests have passed, so functionality-wise it is complete. Next up I'll add doc string to new functions.

masahi:
@interesaaat The doc is done and I think it is ready for final review. Let me know your thought

interesaaat:
Thanks @masahi, I am looking at the PR now.

hummingbird/ml/_container.py
```diff
@@ -81,25 +80,105 @@ def _run(self, function, *inputs, reshape=False):
             splits = [inputs[input_names[idx]] for idx in range(len(input_names))]
             inputs = [df.to_numpy().reshape(-1, 1) for df in splits]
 
-        if self._batch_size is None:
-            return function(*inputs)
-        else:
-            return self._run_batch_inference(function, *inputs, reshape=reshape)
+        return function(*inputs)
```
interesaaat, line 83: Nit: the documentation of this method needs a refresh I think (sorry but I wasn't able to add a comment to the proper line for some reason). `This function either score the full dataset at once or triggers batch inference.` Now it always score the full dataset, right?

hummingbird/ml/_topology.py
```diff
@@ -261,17 +265,15 @@ def fix_graph(graph):
...
 
         tvm_model = _compile_tvm_model(topology, torch_model, batch_trace_input, target, ctx, config, extra_config)
 
         if remainder_trace_input is not None:
-            tvm_remainder_model = _compile_tvm_model(
+            remainder_model = _compile_tvm_model(
                 topology, torch_model, remainder_trace_input, target, ctx, config, extra_config
             )
-            extra_config[constants.TVM_REMAINDER_MODEL] = tvm_remainder_model
+            extra_config[constants.TVM_REMAINDER_MODEL] = remainder_model
```
interesaaat: I don't think we need this anymore. Probably also `constants.TVM_REMAINDER_MODEL` can be removed.

interesaaat (changes requested):

hummingbird/ml/_topology.py
masahi: Good catch removed.

hummingbird/ml/_container.py
masahi: The doc updated. Also removed `reshape=False` option, as it is not used anymore.

interesaaat (approved):

interesaaat:
Thanks for the contribution @masahi!

PR #378, opened 2020-12-04
Fix problem with ONNX models

interesaaat:
This PR fixes few issue with the onnx exporter:
- it fixes the error raised when  inputs of shape == 0 are passed
- it adds proper data types for scalers
- it removes ONNX_INPUTS since is not used

### DIFF
--- a/hummingbird/ml/convert.py
+++ b/hummingbird/ml/convert.py
@@ -231,7 +231,7 @@ def _convert_onnxml(model, backend, test_input, device, extra_config={}):
             extra_config[constants.TEST_INPUT] = test_input
 
     # Set the number of features. Some converter requires to know in advance the number of features.
-    if constants.N_FEATURES not in extra_config:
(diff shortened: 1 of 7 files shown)


hummingbird/ml/convert.py
```diff
@@ -181,6 +181,11 @@ def _convert_onnxml(model, backend, test_input, device, extra_config={}):
                     raise RuntimeError("Cannot fetch input shape from ONNX schema. Please provide some test input.")
                 shape = [dim.dim_value for dim in input.type.tensor_type.shape.dim]
 
+                assert len(shape) == 2
+                # In ONNX dynamic dimensions will have a shape of 0. Fix the 0-shape in the batch dimension if they exist.
+                if shape[0] == 0:
+                    shape[0] = 1
```
ksaur, line 187: add a test for line 187

ksaur (approved):

Issue #389, opened 2020-12-14
Support lightgbm >= 3


memeplex:
setup.py requires lightgbm < 3, I'm using the current version of lightgbm so hummingbird reports that it's not installed. Is there any blocking incompatibility or is it just that the requirement version is out of date?

memeplex:
Not the same, but related https://github.com/microsoft/hummingbird/issues/282

interesaaat:
I think we just didn't test it with >= 3. We found #282 when we tried with v3. As far as I remember this was the only problem, so beside random forest it should just work.

memeplex:
That was fast! It's weird that it is complaining that lightgbm is not installed then, but if you say it should work I will give it another try and let you know, thanks!

interesaaat:
I saw warning that the version is not supported etc., but not real errors. So if the complaining is just a warning, I think it should work. Otherwise if you installed from source, you can just change `setup.py` and remove the `<3` for lightgbm.

memeplex:
Ok, another similar issue is that it wants to downgrade numpy from 1.19.4, do you think it would be a problem if I keep 1.19.4, now that I'm in the business of forcing dependencies?

And even more OT: currently I'm using onnxtools, which is compiling to the tree ensamble operator (in the ML set) that is later evaluated by onnx runtime (C++ code that more or less straightforwardly evals the ensamble). Should I expect a significative performance gain if I compile without the ML set using hummingbird?

memeplex:
Ok, I've been trying again, this was my test code:

```python
from hummingbird.ml import convert
import numpy as np
import lightgbm as lgb
import timeit
 
import onnxruntime as ort
from onnxmltools.convert import convert_lightgbm
from onnxconverter_common.data_types import FloatTensorType

max_depth = 4
n_estimators = 100
n_features = 100
n_classes = 2
n_cats = 20
n_fit = 10000
n_pred = 10
X_fit = np.random.randint(n_cats, size=(n_fit, n_features)).astype(np.float32)
y_fit = np.random.randint(n_classes, size=n_fit)

model = lgb.LGBMClassifier(n_estimators=n_estimators, max_depth=max_depth)
model.fit(X_fit, y_fit, categorical_feature=list(range(n_features)))

input_types = [("input", FloatTensorType((n_pred, n_features)))]
onnx_model = convert_lightgbm(model, initial_types=input_types)
hum_onnx_model = convert(model, 'onnx', extra_config={"initial_types": input_types})
[...]

Issue #372, opened 2020-11-29
Numpy install error


ThomasWolf0701:
I tried to install hummingbird with pip and it tried to downgrade numpy to an old version, and gives an error.

Would it be possible to install via conda ?

WARNING: Error parsing requirements for numpy: [Errno 2] No such file or directory: 'c:\\users\\thomas wolf\\anaconda3\\envs\\my-rdkit-env\\lib\\site-packages\\numpy-1.19.4.dist-info\\METADATA'
Installing collected packages: numpy, onnx, onnxconverter-common, hummingbird-ml
  Attempting uninstall: numpy
    Found existing installation: numpy 1.19.4
ERROR: Could not install packages due to an EnvironmentError: [Errno 2] No such file or directory: 'c:\\users\\thomas wolf\\anaconda3\\envs\\my-rdkit-env\\lib\\site-packages\\numpy-1.19.4.dist-info\\RECORD'

ksaur:
Hi @ThomasWolf0701 ! We don't currently have hummingbird in conda (see #314).

I think the easiest way to install Hummingbird if it conflicts with an existing `numpy` installation would be to create a fresh virtual environment.  Does that work for you?

ksaur:
closed with #390

PR #390, opened 2020-12-14
trying with newer numpy

ksaur:
Investigating newer  numpy issue to avoid downgrade mentioned in #372 and also #389


ksaur:
In the CI/CD  pipeline, `numpy==1.18.5` is installed  on  `python==3.5` and then now with this change  `numpy==1.19.4` is installed on `python>3.5`.

interesaaat (approved):

PR #399, opened 2020-12-22
Add save\load to containers

### LINKED ISSUE #322: Add save\load primitives to Hummingbird
With the current design, saving and loading models has a couple of drawbacks (listed in #321). To overcome these we can add `save` and `load` primitives to HB.

A possibility is to have `save` and `load` as part of the package (as for `torch.save` and `onnx.save`) so that users can do `hummingbird.ml.save` for instance.

We should also have an option to save only the base model and not the container. Something like a `save_native_model_only=False` flag.

interesaaat:
This PR adds a save and load API to the containers.

Closes #322.

### DIFF
--- a/hummingbird/ml/_container.py
+++ b/hummingbird/ml/_container.py
@@ -72,6 +72,12 @@ class SklearnContainer(ABC):
 
     @abstractmethod
     def save(self, location):
+        """
+        Method used to save the container for future use.
+
+        Args:
+            location: The location on the file system where to save the model.
+        """
         return
 
     def _run(self, function, *inputs):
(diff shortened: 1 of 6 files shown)


hummingbird/ml/_container.py
```diff
@@ -319,12 +330,65 @@ def score_samples(self, *inputs):
...
+
+        if constants.TEST_INPUT in self._extra_config:
+            self._extra_config[constants.TEST_INPUT] = None
+
+        if "torch.jit" in str(type(self.model)):
+            assert not os.path.exists(location), "Directory {} already exists.".format(location)
+            os.makedirs(location)
+            self.model.save(os.path.join(location, "deploy_model.zip"))
+            model = self.model
+            self._model = None
```
ksaur, line 350: Is this line necessary?  (it will just get clobbered on 351 anyway right?)

hummingbird/ml/_container.py
```diff
@@ -319,12 +330,65 @@ def score_samples(self, *inputs):
...
+            with open(os.path.join(location, "container.pkl"), "wb") as file:
+                dill.dump(self, file)
+            self._model = model
+        else:
```
ksaur: Should we do additional checking here? Just in case a user tries to save some crazy model?

hummingbird/ml/_container.py
```diff
@@ -319,12 +330,65 @@ def score_samples(self, *inputs):
...
+    def save(self, location):
+        assert self.model is not None, "Saving a None model is undefined."
+
+        if constants.TEST_INPUT in self._extra_config:
+            self._extra_config[constants.TEST_INPUT] = None
+
+        if "torch.jit" in str(type(self.model)):
+            assert not os.path.exists(location), "Directory {} already exists.".format(location)
+            os.makedirs(location)
+            self.model.save(os.path.join(location, "deploy_model.zip"))
```
ksaur: Maybe make "deploy_model.zip" a constant?  Not sure.  I see we have lots of such strings, maybe they are ok as-is....is there some convention?   Or maybe make an optional parameter?

hummingbird/ml/_container.py
```diff
@@ -319,12 +330,65 @@ def score_samples(self, *inputs):
...
+            with open(os.path.join(location, "container.pkl"), "rb") as file:
+                container = dill.load(file)
+            container._model = model
+        else:
+            # This is a pytorch  model
+            with open(location, "rb") as file:
+                container = dill.load(file)
+
+        if container._n_threads is not None:
+            if torch.get_num_interop_threads() != 1:
```
ksaur, line 391: Can you please put a comment here explaining why? :)

ksaur (commented):
This is great!  We should also add documentation to the README maybe?  Or somewhere?

hummingbird/ml/_container.py
interesaaat: Yes this is necessary because otherwise dill fails. Basically I need to remove from the container all references to objects that cannot be dumped. No idea if there is a better way of doing this.

hummingbird/ml/_container.py
interesaaat: fixed.

hummingbird/ml/_container.py
interesaaat: Yep. Fixed.

hummingbird/ml/_container.py
interesaaat: Fixed.

PR #401, opened 2020-12-28
Better load model support

interesaaat:
This PR fixes a couple of bug triggered by the latest version of TVM when saving \ loading a model.

Additionally, it allows to easy load a model without having to specific the container type.

### DIFF
--- a/hummingbird/ml/_utils.py
+++ b/hummingbird/ml/_utils.py
@@ -11,6 +11,7 @@ Collection of utility functions used throughout Hummingbird.
 from distutils.version import LooseVersion
 import numpy as np
 import torch
+import os
 import warnings
 
 from hummingbird.ml.exceptions import ConstantError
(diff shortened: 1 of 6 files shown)


hummingbird/ml/_utils.py
```diff
@@ -180,6 +180,47 @@ def from_strings_to_ints(input, max_string_length):
...
+    # Add load capabilities
+    from ._container import PyTorchSklearnContainer
+    from ._container import TVMSklearnContainer
+    from ._container import ONNXSklearnContainer
+
+    model = None
+
+    # Try as TVM model.
+    try:
+        model = TVMSklearnContainer.load(location)
```
ksaur: I wondered why you were doing it this way when I learned pickle has[ no header](https://stackoverflow.com/questions/13939913/how-to-test-if-a-file-has-been-created-by-pickle)?  This is clunky but I don't see a good way around.

ksaur (approved):

hummingbird/ml/_utils.py
interesaaat: Fixed!

PR #408, opened 2020-12-30
Fix scaling in onnx

interesaaat:


hummingbird/ml/operator_converters/onnx/scaler.py
```diff
@@ -32,9 +32,9 @@ def convert_onnx_scaler(operator, device=None, extra_config={}):
 
     for attr in operator.origin.attribute:
         if attr.name == "offset":
-            offset = np.array(attr.floats).astype("float32")
+            offset = np.array(attr.floats)
```
ksaur, line 35: What helped find this bug? Should we add a different type of data on `test_onnxml_scaler_converter.py` line 24?

interesaaat: I had some problem with accuracy on an onnx model. Yes let me add a test with float64.

PR #409, opened 2020-12-30
Refactoring of topology

interesaaat:


hummingbird/ml/_model_runner.py
```diff
@@ -0,0 +1,121 @@
...
+from hummingbird.ml._utils import pandas_installed, get_device, from_strings_to_ints
+from hummingbird.ml.operator_converters import constants
+
+if pandas_installed():
+    from pandas import DataFrame
+else:
+    DataFrame = None
+
+
+class ModelRunner(torch.nn.Module, object):
```
ksaur: why the name "Runner"?  Runner makes me think of the constant tests in CI/CD systems, but maybe that's just me.  It's an internal representation right...maybe "base"? i dunno

ksaur (approved):

hummingbird/ml/_model_runner.py
interesaaat: I was divided between Runner and Executor. Do you prefer Executor? Basically it is the class "running" the converted pipeline.

ksaur: Yah i think Executor maybe?  I don't feel strongly.  I think I'm biased against `runner` based on some unrelated ci/cd testing work i did, but I don't think it's a universal term.  It's up to you :)

PR #410, opened 2020-12-30
Add padding to TVM

interesaaat:
TVM by default statically compile models. This basically means that we need to set the input shape at conversion time and if the input shape of the prediction data do not match, TVM fails. 

This PR fixes this problem by allowing to pad input data at prediction time to match the size of the test data used at conversion time.


ksaur:
Ugh, switching the `base` like i just did used to auto-rebase the PR for review.  now it has all the refactorings in it. :-P =

hummingbird/ml/containers/sklearn/tvm_containers.py
```diff
@@ -176,17 +182,31 @@ def load_param_dict(param_bytes):
...
+        # Compute padding.
+        padding_size = 0
+        if self._pad_input:
+            padding_size = self._batch_size - inputs[0].shape[0]
+
+        # Prepare inputs.
+        for i, input_ in enumerate(inputs):
+            assert (
+                self._pad_input or input_.shape[0] == self._batch_size
+            ), "The number of input rows {} is different from the batch size {} the TVM model is compiled for.".format(
```
ksaur, line 197: In what case would this fail?  If you're doing `padding_size = self._batch_size - inputs[0].shape[0]` above, then isn't this guaranteed to be true.   (Assuming not garbage input I guess?)

ksaur (approved):

hummingbird/ml/containers/sklearn/tvm_containers.py
interesaaat: It should fail when pad is false and `input_.shape[0]` and `self._batch_size` are different, no?

ksaur: Ah!  now i see that TVM_PAD_PREDICTION_INPUT can be false, got it

masahi:
This is ok, but not encouraged. This kills performance if padding is large. If the input size can vary widely, TVM is not a good fit for that use case.

interesaaat:
> This is ok, but not encouraged. This kills performance if padding is large. If the input size can vary widely, TVM is not a good fit for that use case.

Agreed. It would be good to have something like TVM VM (with good performance).

PR #415, opened 2021-01-04
Add support for modified huber loss in SGD

interesaaat:
Closed #413.

PR #416, opened 2021-01-05
Fix typo

interesaaat:


interesaaat:
Mmm I just realized it has the same code from #415, so let's merge it after that one.

PR #387, opened 2020-12-12
trying to get pipeline working

ksaur:
LGBM is failing on Macos (not python3.5 apparently)... seemingly due to some new update today

```
tests/test_onnxml_lightgbm_converter.py:124: 

self = <test_onnxml_lightgbm_converter.TestONNXLightGBMConverter testMethod=test_lgbm_onnxml_model_regressor>
X = array([[0.5488135 , 0.71518934, 0.60276335, ..., 0.639921  , 0.14335328,
        0.9446689 ],
       [0.5218483 , 0.41...5576258],
       [0.28058997, 0.38613838, 0.2736383 , ..., 0.73407567, 0.8579786 ,
        0.45330405]], dtype=float32)
model = LGBMRegressor(boosting_type='gbdt', class_weight=None, colsample_bytree=1.0,
              importance_type='split', le...e, reg_alpha=0.0, reg_lambda=0.0, silent=True,
              subsample=1.0, subsample_for_bin=200000, subsample_freq=0)
rtol = 1e-06, atol = 1e-06, extra_config = {}

    def _test_regressor(self, X, model, rtol=1e-06, atol=1e-06, extra_config={}):
        onnx_ml_pred, onnx_pred, output_names = self._test_lgbm(X, model, extra_config)
    
        # Check that predicted values match
>       np.testing.assert_allclose(onnx_ml_pred[0], onnx_pred[0], rtol=rtol, atol=atol)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-06, atol=1e-06
E       
E       Mismatched elements: 1 / 100 (1%)
E       Max absolute difference: 3.8146973e-05
E       Max relative difference: 1.1161203e-06
E        x: array([[32.70483 ],
E              [35.15778 ],
E              [83.108055],...
E        y: array([[32.704826],
E              [35.157776],
E              [83.10807 ],...
```

Pinning to an older macos didnn't fix it.  Setting rtol./atol for now

PR #418, opened 2021-01-05
updating macos ci/cd

ksaur:
seeing if macos-10.15 works now (previously changed from -latest due to unexpected version conflict with onnxml lgbm  #387)


ksaur:
It's unclear to me if this actually shaves a few minutes off the ci/cd runtime, but it will definitely make the warning`(The macOS virtual environment has been updated to Catalina (v10.15). Please update your workflow and change the line 'runs-on: macOS-10.14`) go away.  
We still have the weirdness with python3.5+mac v10.15, but we handled that with `rtol`.

interesaaat (approved):

PR #428, opened 2021-01-11
Remove duplicated code

qin-xiong:
The code for getting n_classes in method get_parameters_for_gemm_common is repeated twice. One should be removed.


ksaur:
Thanks for this @qin-xiong !

PR #429, opened 2021-01-11
Fix bloated models when saving

interesaaat:
For some reason we were storing the full topology in each Executor. This bloats the models when we save them. 

On my local machine this PR creates ~3x smaller models.

Unfortunately I had to change all converters to propagate the change :/


hummingbird/ml/operator_converters/sklearn/decomposition.py
```diff
@@ -47,6 +49,8 @@ def convert_sklearn_kernel_pca(operator, device, extra_config):
     Returns:
         A PyTorch model
     """
+    assert operator is not None
```
ksaur: I notice we have this `assert` at the start of each converter.  In what case does this fail?  If the model is corrupted?  (Is there some error message that we could give users that would be helpful?)

ksaur (approved):
Thanks, this was much easier to read/check with the reformat. :)

hummingbird/ml/operator_converters/sklearn/decomposition.py
interesaaat: I think this is hard to trigger. I believe it will fail before actually reaching this point. I added them as a double check. I will add an error message. Something like "cannot convert None operator".

PR #435, opened 2021-01-29
Set missing direction to left for trees with single leaf to avoid missing value checks in compiled code

### LINKED ISSUE #434: [-1] is treated as "non trivial missing value"
The following trivial program using TVM backend doesn't finish compiling, with the latest main.
```
import numpy as np
from sklearn.datasets import load_iris

import xgboost as xgb
import hummingbird.ml


X, y = load_iris(return_X_y=True)
X = X.astype(np.float32)  # make sure to use fp32 input

model = xgb.XGBClassifier(
    n_estimators=100,
    max_depth=8,
    learning_rate=0.1,
    reg_lambda=1,
    objective="multi:softmax",
    num_class=3
)
model.fit(X, y)

tvm_model = hummingbird.ml.convert(model, "tvm", X)
np.testing.assert_allclose(model.predict(X), tvm_model.predict(X))
```

I did bisect and found that after the commit https://github.com/microsoft/hummingbird/pull/419 that added missing value support, the infinite compilation time began to happen.

I don't know what's causing TVM troubles, but one thing I found was that after https://github.com/microsoft/hummingbird/pull/419, `isnan` appears in my graph, even though there is no missing values in the dataset. 

After some debugging, I found that this line https://github.com/scnakandala/hummingbird/blob/57dfbc9ec91813a562b53c4208ed3d1a132df2c6/hummingbird/ml/operator_converters/_tree_commons.py#L212 would hit, where in my case `missings` is [-1]. Since [-1] is not None, this if becomes True, and the rest of the conversion thinks this is a missing value.

-1 is introduced in https://github.com/scnakandala/hummingbird/blob/57dfbc9ec91813a562b53c4208ed3d1a132df2c6/hummingbird/ml/operator_converters/xgb.py#L31. I think [-1] should not be treated as a missing value.

Adding the following fixed it for me:
```
        if missings == [-1]:
            missings = None
```
@scnakandala @interesaaat

scnakandala:
closes #434

### DIFF
--- a/hummingbird/ml/operator_converters/_tree_commons.py
+++ b/hummingbird/ml/operator_converters/_tree_commons.py
@@ -210,7 +210,7 @@ def get_parameters_for_tree_trav_common(lefts, rights, features, thresholds, val
         features = [0, 0, 0]
         thresholds = [0, 0, 0]
         if missings is not None:
-            missings = [2, -1, -1]
+            missings = [1, -1, -1]
         n_classes = values.shape[1] if type(values) is np.ndarray else 1
         values = np.array([np.zeros(n_classes), values[0], values[0]])
         values.reshape(3, n_classes)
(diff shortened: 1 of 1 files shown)


masahi (approved):
I confirmed it fixed the issue, thanks!

masahi:
So was this a real bug? I don't understand the fix.

scnakandala:
Yes, it was a bug. It happens for a corner case where we have tree with only one leaf and TVM does not support is_nan.
Earlier, for a tree with one leaf node we chose the right side (i.e., 2 - True side) as the default. Now I have changed it to left side (1 - False side). 

In the tree implementations, we check whether the model requires non-trivial missing value handling. If all missing values take the left side (False side) we don't have to explicitly handle missing values. A comparison with nan is always False. So the model will not need is_nan.

PR #441, opened 2021-02-11
Deprecate Python 3.5

ksaur:
Holding off on adding Python 3.9 due to ONNX dependencies not yet ready

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, macos-10.15, windows-2019]
-        python-version: [3.5, 3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/pythonapp.yml
```diff
@@ -36,19 +36,19 @@ jobs:
         # The GitHub editor is 127 chars wide
         flake8 . --count  --max-complexity=10 --max-line-length=127 --statistics
     # PyTorch for Mac has different pip syntax wrt Win and Linux.
-    # PyTorch stop supporting python 3.5 from version 1.6.
+    # We want to test torch 1.5.x (in this case on python3.6) and torch 1.7.x.
     # The following cases address the situations above.
-    - name: Install pytorch 1.5.1 if python 3.5 (mac)
-      if: ${{ matrix.python-version == '3.5' && matrix.os == 'macos-10.15' }}
+    - name: Install pytorch 1.5.1 if python 3.6 (mac)
```
interesaaat, line 41: 1.5.1 only supports 3.6? I think we can get rid of this special case and run all pytorch versions on 3.6 >=

.github/workflows/pythonapp.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         pip install pandas
     - uses: actions/cache@v1
       # TVM takes forever, we try to cache it.
-      if: ${{ matrix.python-version != '3.5' && matrix.os != 'windows-2019' }}
+      if: ${{ matrix.python-version != '3.6' && matrix.os != 'windows-2019' }}
```
interesaaat: Since we removed 3.5, I don't think we need to have special paths for 3.6 and 3.7. We can just use 3.7. This should also speed up the pipeline hopefully

.github/workflows/pythonapp.yml
ksaur: Before, we had Python 3.5 on `torch==1.5.1` and all other versions of python on `torch==1.7.1`.  When we removed python 3.5, we still want to keep the old torch, so I slid python3.6 into that slot.

Do you mean to have python 3.6 run twice, once for old and once for new torch? I'm not quite sure how to set that up cleanly with the 'matrix' we have now, but it's possible.

.github/workflows/pythonapp.yml
ksaur: ok removing that part!

.github/workflows/pythonapp.yml
ksaur: There are two code paths (with `LooseVersion`, ex:`@unittest.skipIf(LooseVersion(torch.__version__) < LooseVersion("1.6.0")`,).  I think maybe this is only for SparkML?

Maybe we are ok with testing only `torch==1.7.1`?  I am looking at the coverage tool right now to see what changes, but it is not cooperating!

interesaaat: Ah I see now. And running all python versions for 2 versions of pytorch will be bad for our pipeline. Then it is ok!

ksaur: Ok cool. :) I am still looking into why coverage says we drop 6%. I think it's just an error...going to the actual coverage link i *think* it is ok.

interesaaat: I think that the drop is because only 3.6 finishes and many tests are skipped, meaning less coverage. I think we had similar problems for 3.5 when only those runs were finishing.

ksaur: I thought they fixed that so that it refreshed but it seems no.   

Looks all good now! :) https://codecov.io/gh/microsoft/hummingbird/pull/441/changes

PR #449, opened 2021-02-24
Fixing test flakiness

sleepy-owl:
The test `test_sklearn_multioutput_regressor` sometimes fails. This PR address this issue.

To find a solution, I collected samples of the differences in the values being compared in the assertion, from several test executions and computed the tail distribution. I computed the extreme percentiles to check how high can the differences be. 
These are the percentiles for the relative differences:

```
0.9 : 5.8e-5
0.99: 7.6e-5
0.999: 5e-5
0.9999: 1e-4
```

For this fix, I chose the 99.99th percentile. I think setting the bound using the statistical evaluation might be a good way to ensure the test is not flaky.

Do you guys think this makes sense? Please let me know if this looks good or if you have any other suggestions. Also, here I assume there are no bugs in the code under test.

### DIFF
--- a/tests/test_sklearn_multioutput_regression.py
+++ b/tests/test_sklearn_multioutput_regression.py
@@ -35,7 +35,7 @@ class TestSklearnMultioutputRegressor(unittest.TestCase):
 
                 torch_model = hummingbird.ml.convert(model, "torch")
                 self.assertTrue(torch_model is not None)
-                np.testing.assert_allclose(model.predict(X), torch_model.predict(X), rtol=1e-5, atol=1e-5)
+                np.testing.assert_allclose(model.predict(X), torch_model.predict(X), rtol=1e-4, atol=1e-5)
 
     # Test RegressorChain with different child learners
     def test_sklearn_regressor_chain(self):


interesaaat:
Thanks for checking this. I think it makes sense. We have some loss in accuracy for some operator due to rounding errors. `1e-4` is still good.

sleepy-owl:
Thanks @interesaaat! I will also be happy to look into any other flaky tests that you may know of. I can try fixing them as well.

ksaur:
Just curious, how often did you see failures? We haven't seen it fail recently in the ci/cd, so I'm curious what your experience was in terms of failure rate?

sleepy-owl:
It failed 5 out of 500 times for me.

interesaaat:
I think we have other tests having a similar problem. I need to check which one. For the moment we manually set a "reasonable" value for the differences, but having a more "scientific" way of finding them will be great!

sleepy-owl:
@interesaaat @ksaur  Just curious, is there some randomness involved in how hummingbird converts a scikit-learn model? I wonder why the "loss in accuracy" and "rounding errors" occur and why is it different across executions?

interesaaat:
I think the randomness here is into the dataset. `datasets.make_regression` generates a random dataset. The conversion and evaluation code is deterministic AFAIK.

sleepy-owl:
Oh, I see, thanks!

sleepy-owl:
Also, are there any other reasons why the predictions from the two models could differ (apart from rounding errors from operators)?

Do you know which operators are more likely to introduce rounding errors?

PR #169, opened 2020-06-25
Make GEMMTreeImpl strictly follows scikit-learn's behavior at decision boundary

gyeongin:
Closes #30.

PR #450, opened 2021-02-25
Fixing test flakiness

sleepy-owl:
Hi,

The test  `test_tree_regressors_multioutput_regression` is flaky. It failed 41/3000 times that I ran. It seems the absolute difference can be much higher than the current threshold (1e-5).

Empirically, I observed a value of a maximum absolute difference of upto 3.7. Based on my experiments, the 99th percentile seems to be close to 4.5. Hence, I set this bound accordingly. 

Please let me know if this makes sense. I am assuming there are no bugs in this case. I would be happy to incorporate any other suggestions that you may have.

Thanks!

### DIFF
--- a/tests/test_sklearn_decision_tree_converter.py
+++ b/tests/test_sklearn_decision_tree_converter.py
@@ -725,7 +725,7 @@ class TestSklearnTreeConverter(unittest.TestCase):
 
                     torch_model = hummingbird.ml.convert(model, "torch", extra_config={constants.TREE_IMPLEMENTATION: tree_method})
                     self.assertTrue(torch_model is not None)
-                    np.testing.assert_allclose(model.predict(X), torch_model.predict(X), rtol=1e-5, atol=1e-5)
+                    np.testing.assert_allclose(model.predict(X), torch_model.predict(X), rtol=1e-5, atol=4.5)
 
 
 if __name__ == "__main__":


interesaaat:
Is this `1e-4.5` or `4.5`? `4.5` is quite high.

sleepy-owl:
It is 4.5 actually. In several cases the absolute difference exceeds 3. Do you think there is a possible bug? Or should the test be configured differently (such as more samples)?

interesaaat:
4.5 is quite large, I am now thinking that maybe there is a bug somewhere? @scnakandala any idea?

scnakandala:
4.5 seems a bug. But it is hard to see where it is coming from. I will check this.

sleepy-owl:
do you know if this might be a bug in code or test?

interesaaat:
I think that testing is ok, if it's a bug it is probably in the code.

interesaaat:
@scnakandala any news on this?

ksaur:
Should we move this to a bug report rather than a PR?  I don't think it's acceptable to have `atol=4.5` in our tests!

scnakandala:
I did some digging into this and I think the solution is to revive @gyeongin's previous pull request [1]. Basically, the problem seems to be GEMM tree implementation having a `<` condition instead of `<=` condition.

[1] - https://github.com/microsoft/hummingbird/pull/169

interesaaat:
I see, and the other tree implementation were already using `<=` right? I am surprised this is popping up only now.

interesaaat:
If I change to `<=` now we have [failures in LightGBM with TVM](https://github.com/microsoft/hummingbird/pull/169/checks?check_run_id=2452649019).

scnakandala:
[This test](https://github.com/microsoft/hummingbird/pull/169/checks?check_run_id=2452649019)'s GEMM fails even with torch backend. Comparing only the *.predict_proba(X) I found that the differences are in the order of 10^-3 to 10^-2.

I think something more is happening here. I will check more and update.

interesaaat:
You it is failing with `<=` also for torch, right? Not that it is failing if we have just `<`.

interesaaat:
Fixed by #511.

PR #451, opened 2021-02-25
Fixing test flakiness

sleepy-owl:
Hi,

The test `test_sklearn_regressor_chain` is flaky. It failed 130/1050 times that I ran. It seems that the absolute difference between the prediction can often exceed 1e-5. In one case, it went up to 5.33. I computed the extreme percentiles from the tail distribution.

```
0.9 :: 0.017
0.99 :: 1.65 ~ 1.7
0.999 :: 5.16
0.9999 :: 5.32
```

I used the 99th percentile in this case. Do you guys think this makes sense? Should I use a different percentile? I am assuming there are no bugs in the code/test in this case. 

Thanks!


interesaaat:
Fixed by #511.

PR #456, opened 2021-03-05
onnx svc

ksaur:
SVC converter

SKL->torch working for binary/multiclass.  (Does not work for multiclass TVM due to bug)
ONNXML->torch working for binary/multiclass.
ONNXML->onnx has a bug, documented in the testcase (GEMM not implemented in onnx converter)


ksaur:
Ok this hits everything except `ArgMax` and `MatMul` which was expected.  Still a ~0.43% hit...slightly more than i thought it would be.

hummingbird/ml/operator_converters/onnx/onnx_operator.py
```diff
@@ -64,15 +64,18 @@ class Sum(PhysicalOperator, torch.nn.Module):
...
+            x = torch.cat(x, dim=1)
+        return torch.sum(*x)
 
 
 class Add(PhysicalOperator, torch.nn.Module):
     def __init__(self, logical_operator, val):
         super(Add, self).__init__(logical_operator)
 
-        self.val = torch.nn.Parameter(torch.FloatTensor(val), requires_grad=False)
+        if val is not None:
```
interesaaat, line 77: Can we have an assert here that checks that `len(self.inputs) == 1` when `val is not None`? We need to make sure that indeed it will go in the right branch inside `forward`

hummingbird/ml/operator_converters/onnx/onnx_operator.py
```diff
@@ -110,10 +113,13 @@ class Mul(PhysicalOperator, torch.nn.Module):
     def __init__(self, logical_operator, val):
         super(Mul, self).__init__(logical_operator)
 
-        self.val = val
+        if val is not None:
```
interesaaat, line 117: Same here.

hummingbird/ml/operator_converters/onnx/onnx_operator.py
```diff
@@ -256,7 +262,10 @@ def convert_onnx_add(operator, device=None, extra_config={}):
     assert operator is not None
 
     initializers = extra_config[constants.ONNX_INITIALIZERS]
-    val = list(initializers[operator.raw_operator.origin.input[1]].float_data)
+    if operator.raw_operator.origin.input[1] not in initializers:
```
interesaaat: Can we do like:
`val = None`
`if operator.raw_operator.origin.input[1] in initializers:`
bla
(basically avoid the if\else and use `val = None` as default case)

hummingbird/ml/operator_converters/onnx/onnx_operator.py
```diff
@@ -313,7 +322,10 @@ def convert_onnx_mul(operator, device=None, extra_config={}):
...
```
interesaaat: Same here

hummingbird/ml/operator_converters/onnx/sv.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Args:
+        operator: An operator wrapping a `ai.onnx.ml.SVMClassifier` model
+        device: String defining the type of device the converted operator should be run on
+        extra_config: Extra configuration used to select the best conversion strategy
+
+    Returns:
+        A PyTorch model
+    """
+    operator_orig = operator
```
interesaaat: Can we avoid this and just call operator.raw_operator (instead of operator) when required later in the code (apparently only in line 38)

hummingbird/ml/operator_converters/onnx/sv.py
```diff
@@ -0,0 +1,96 @@
...
+            degree = int(kp_arr[2])
+
+        elif attr.name == "classlabels_ints":
+            classes = np.array(attr.ints)
+
+    if any(v is None for v in [sv_vals, coeffis]):
+        raise RuntimeError("Error parsing SVC arrays, found unexpected None")
+
+    # Now that we have parsed the degree and lengths, reshape 'a' and 'sv'
+    #   For 'a', these are in 'dual' shape, so resize into 2:
```
interesaaat: Remove spacing here please.

tests/test_onnxml_sv_converter.py
```diff
@@ -0,0 +1,112 @@
...
+        onnx_ml_pred = [[] for i in range(len(output_names))]
+        inputs = {session.get_inputs()[0].name: X}
+        pred = session.run(output_names, inputs)
+        for i in range(len(output_names)):
+            if "label" in output_names[i]:
+                onnx_ml_pred[1] = pred[i]
+            else:
+                onnx_ml_pred[0] = pred[i]
+
+        if mode == "torch":
```
interesaaat: I don't think we need this if, with the containers now we do the same thing between onnx and torch models.

interesaaat (changes requested):

tests/test_onnxml_sv_converter.py
ksaur: I can pull out `convert` and pass the mode as an arg, but lines 58 (`pred = torch_model.predict(X)`) and `pred = [[] for i in range(len(output_names))]`...etc are different, unless you see a clean way to combine them?

hummingbird/ml/operator_converters/onnx/onnx_operator.py
ksaur: good idea!

tests/test_onnxml_sv_converter.py
ksaur: oh wait, you're right, we don't need that. let me fix

interesaaat (approved):

PR #459, opened 2021-03-08
onnxml imputer

ksaur:
Adding Imputer for Onnxml.

It seems that the existing converter for Imputer relied on `raw_operator` being SKL, so i parameterized the constructor to fit.


hummingbird/ml/operator_converters/_imputer_implementations.py
```diff
@@ -0,0 +1,81 @@
...
+    Class implementing SimpleImputer operators in PyTorch.
+    """
+
+    def __init__(self, logical_operator, device, statistics=None, missing=None, strategy=None):
+        super(SimpleImputer, self).__init__(logical_operator)
+        sklearn_imputer = logical_operator.raw_operator
+        # Pull out the stats field from either the SKL imputer or args
+        stats_ = statistics if statistics is not None else sklearn_imputer.statistics_
+        # Process the stats into an array
+        stats = [float(stat) for stat in stats_ if isinstance(stat, float)]
```
interesaaat: Why do you need to check the type here? I think we can skip this and just use straight stats?

interesaaat: (I know that this is not your code)

hummingbird/ml/operator_converters/onnx/imputer.py
```diff
@@ -0,0 +1,45 @@
...
+        if attr.name == "imputed_value_floats":
+            stats = np.array(attr.floats).astype("float64")
+        elif attr.name == "replaced_value_float":
+            missing = attr.f
+
+    if any(v is None for v in [stats, missing]):
+        print("stats: {}, missing: {}".format(stats, missing))
+        raise RuntimeError("Error parsing Imputer, found unexpected None")
+
+    # ONNXML has no "strategy" field, so pass the default SKL field, which is "mean"
```
interesaaat: I am not sure this is correct. I think that their behavior is changing the value with a constant. So I believe the strategy is actually "constant".

tests/test_onnxml_imputer_converter.py
```diff
@@ -0,0 +1,89 @@
...
+    import onnxruntime as ort
+if onnx_ml_tools_installed():
+    from onnxmltools import convert_sklearn
+    from onnxmltools.convert.common.data_types import FloatTensorType as FloatTensorType_onnx
+
+
+class TestONNXImputer(unittest.TestCase):
+    def _test_imputer_converter(self, norm):
+        warnings.filterwarnings("ignore")
+        model = SimpleImputer(strategy="mean", fill_value="nan")
```
interesaaat: Does the tests work only with this strategy? Can we have tests for other strategies? If it is easy to do.

hummingbird/ml/operator_converters/onnx/imputer.py
ksaur: Ah i see. I went with the default for SKL, but you're right based on https://github.com/onnx/onnx/blob/master/docs/Operators-ml.md#aionnxmlimputer it seems that "constant" is the way to go.

tests/test_onnxml_imputer_converter.py
ksaur: I'm not sure if that makes sense/works....the ONNX spec itself doesn't allow you specify. As above, we discussed that onnx uses "constant".  So...to create different SKL models with others should only make sense for tests...or is there a case that I'm missing?

ksaur: it _does_ work for both `model = SimpleImputer(strategy="mean", fill_value="nan")` and `model = SimpleImputer(strategy="constant")` because that field is simply ignored by the converter :)

tests/test_onnxml_imputer_converter.py
```diff
@@ -21,7 +21,7 @@
 class TestONNXImputer(unittest.TestCase):
     def _test_imputer_converter(self, norm):
         warnings.filterwarnings("ignore")
-        model = SimpleImputer(strategy="mean", fill_value="nan")
+        model = SimpleImputer(strategy="constant")
```
interesaaat: Does this mean that onnx imputation supports only constant? It looks then that they somehow support both mean and constants right? Should we have different tests with different modalities for the `SimpleImputer`?

ksaur: Yes it seems so (https://github.com/onnx/onnx/blob/master/docs/Operators-ml.md#aionnxmlimputer)

Added a test case for all cases where they don't throw unsupported (https://github.com/onnx/sklearn-onnx/blob/master/skl2onnx/operator_converters/imputer_op.py)

tests/test_onnxml_imputer_converter.py
```diff
@@ -0,0 +1,110 @@
...
+
+if onnx_runtime_installed():
+    import onnxruntime as ort
+if onnx_ml_tools_installed():
+    from onnxmltools import convert_sklearn
+    from onnxmltools.convert.common.data_types import FloatTensorType as FloatTensorType_onnx
+
+
+class TestONNXImputer(unittest.TestCase):
+    def _test_imputer_converter(self, norm, model=None):
```
interesaaat: I think `norm` is never used.

interesaaat: I don't see how 'l1', 'l2' and 'max' are used. Perhaps you forgot to commit something?

ksaur: Thanks for catching that.  Those are a copy paste error from `normalizer`

interesaaat (approved):

PR #466, opened 2021-03-12
newer numpy

ksaur:


setup.py
```diff
@@ -22,7 +22,7 @@
         long_description = long_description[start_pos:]
 
 install_requires = [
-    "numpy>=1.15,<=1.19.4",
+    "numpy>=1.15,<=1.20.*",
```
interesaaat, line 25: Are we sure we want to go with star? If they update this we. might break.

interesaaat (approved):

setup.py
ksaur: I debated this.  Looking at their [releases](https://pypi.org/project/numpy/#history)  it seems minor releases arerelatively frequent, and in their [release notes](https://github.com/numpy/numpy/releases) they label as "quick release", "bug fix", "small maintenace [sic] release", but I agree it could be a minor risk.  If we do `20.*` we'll have to check it ~6months instead of ~1month.  

I can see it either way.  Up to you! I can change it to `*` to be safer.

interesaaat: Ok I think it is fine, I agree that if they only do small fixes it should be ok.

PR #469, opened 2021-03-19
confirming that torch 1.7 closes issue 10

### LINKED ISSUE #10: Tree classifiers return wrong labels while probabilities match
Apparently `torch.argmax` returns the last matching value, while `np.argmax` return the first one. This is general is not a problem, except when we get the same probabilities for 2 or more classes. 

This will hopefully get fixed once [this ](https://github.com/pytorch/pytorch/issues/10370) will get closed.

Once this is fixed we should add assertion over the labels results as well. At this point this assertions won't be useful since they will likely often fails.

ksaur:
closes #10


ksaur:
Right now our pipeline tests `torch==1.5.1` for python 3.6, and `torch==1.8.0` for python3.7/3.8

interesaaat (approved):

PR #471, opened 2021-03-24
Add support for stack classifier and regressor

### LINKED ISSUE #468: Add 'sklearn.ensemble._stacking.StackingClassifier'
A user requested us to add `sklearn.ensemble._stacking.StackingClassifier`.  (See [doc](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.StackingClassifier.html)).  

At first glance this appears it should be reasonably easy to implement/similar to our other classifier converters.

interesaaat:
Closes #468


hummingbird/ml/_parse.py
```diff
@@ -27,6 +27,11 @@
 from .operator_converters import constants
 from .supported import get_sklearn_api_operator_name, get_onnxml_api_operator_name, get_sparkml_api_operator_name
 
+try:
+    from sklearn.ensemble import StackingClassifier, StackingRegressor
+except ImportError:
+    StackingClassifier = None
```
ksaur, line 34: Can you please put some comment like below on "StackingClassifier not available in scikit-learn < 0.22" so we dont' forget later when we hopefully clean this

tests/test_sklearn_pipeline.py
```diff
@@ -471,7 +478,6 @@ def test_pipeline_column_transformer_drop_noweights(self):
             model.predict_proba(X_test), torch_model.predict_proba(X_test.values), rtol=1e-06, atol=1e-06,
         )
 
-    @unittest.skipIf(ColumnTransformer is None, reason="ColumnTransformer not available in 0.19")
```
ksaur, line 474: Did you pull this b/c  it's now part of `_parse_sklearn_stacking`?  or do these tests just happen to not hit it?

interesaaat: We only support Sklearn >= 0.21.3 so this is redundant.

hummingbird/ml/_parse.py
interesaaat: yep.

PR #490, opened 2021-04-12
Add support for bagging

### LINKED ISSUE #489: please add support for  BaggingClassifier and the base estimator is SVC
SVC is so slow in big dataset，so we can speed up it by using BaggingClassifier，but the model predict is so slow ，please add support for  BaggingClassifier and the base estimator is SVC

interesaaat:
Closes #489.

### DIFF
--- a/tests/test_sklearn_bagging.py
+++ b/tests/test_sklearn_bagging.py
@@ -20,9 +20,6 @@ class TestSklearnBagging(unittest.TestCase):
         hb_model = hummingbird.ml.convert(clf, "torch")
 
         np.testing.assert_allclose(
-            clf.predict([[0, 0, 0, 0]]), hb_model.predict(np.array([[0, 0, 0, 0]])), rtol=1e-06, atol=1e-06,
-        )
-        np.testing.assert_allclose(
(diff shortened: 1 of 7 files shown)


tests/test_sklearn_bagging.py
```diff
@@ -71,7 +71,7 @@ def test_bagging_linear_svr(self):
         hb_model = hummingbird.ml.convert(reg, "torch")
 
         np.testing.assert_allclose(
-            reg.predict(X), hb_model.predict(X), rtol=1e-05, atol=1e-05,
+            reg.predict(X), hb_model.predict(X), rtol=1e-04, atol=1e-04,
```
ksaur, line 74: Let's document this somewhere; maybe in pydoc?

ksaur (approved):
once [bintray](https://status.bintray.com/) comes back up, hopefully the pipeline works again

tests/test_sklearn_bagging.py
interesaaat: I added a comment on `_parse`. Do you think that's enough?

PR #500, opened 2021-04-22
macOS issue with  python3.6 in pipeline

ksaur:
The pipeline surprise-broke today.  After a lot of troubleshooting, it seems it was due to `libomp` version change.  12.0.0 only works with python 3.8 it seems (3.6 and 3.7 broke).  

With some struggles, pinning this to an older (11.1.0) version of libomp. Hopefully there will be a better fix later.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -72,7 +72,7 @@ jobs:
       run: |
         pip install .[extra,onnx,sparkml]
         pip install pandas
-    - uses: actions/cache@v1
+    - uses: actions/cache@v2
       # TVM takes forever, we try to cache it.
       if: ${{ matrix.os != 'windows-2019' }}
       id: cache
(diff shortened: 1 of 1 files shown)


ksaur:
They changed the libomp version from 11 to 12 recently; it seems that is the issue.

But it seems only the latest libomp is in brew, maybe we can use something like https://github.com/dotnet/machinelearning/pull/3721/files

ksaur:
It should work to install this from a hash, but apparently there is an open issue https://github.com/dotnet/docs/issues/21225 and that "It actually looks like with the newer versions of homebrew that you can't use a github url to install from anymore"

ksaur:
Umm...apparently they are not fixing it (installing from github), and everyone is very angry at each other. https://github.com/Homebrew/brew/issues/8791

um.....stepping away slowly....  And trying the work around hopefully....?

interesaaat (approved):

PR #502, opened 2021-04-23
Remove temp dirs when loading / saving models

### LINKED ISSUE #501: Unzipping file twice breaks
Thanks to @dbanda for pointing out this issue:

When you call `load` twice in the same directory, it gives an error
```
    model = load(name)
  File "/home/azureuser/hummingbird/hummingbird/ml/_utils.py", line 203, in load
    shutil.unpack_archive(zip_location, location, format="zip")
  File "/home/azureuser/miniconda3/envs/raven/lib/python3.7/shutil.py", line 993, in unpack_archive
    func(filename, extract_dir, **dict(format_info[2]))
  File "/home/azureuser/miniconda3/envs/raven/lib/python3.7/shutil.py", line 914, in _unpack_zipfile
    _ensure_directory(target)
  File "/home/azureuser/miniconda3/envs/raven/lib/python3.7/shutil.py", line 891, in _ensure_directory
    os.makedirs(dirname)
  File "/home/azureuser/miniconda3/envs/raven/lib/python3.7/os.py", line 223, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/home/azureuser/data/models/DT'

```

interesaaat:
Closes #501.

### DIFF
--- a/hummingbird/ml/containers/sklearn/pytorch_containers.py
+++ b/hummingbird/ml/containers/sklearn/pytorch_containers.py
@@ -80,6 +80,7 @@ class PyTorchSklearnContainer(SklearnContainer):
             with open(os.path.join(location, constants.SAVE_LOAD_TORCH_JIT_PATH), "wb") as file:
                 dill.dump(self, file)
         else:
+            shutil.rmtree(location)
             raise RuntimeError("Model type {} not recognized.".format(type(self.model)))
 
         # Zip the dir.
@@ -128,6 +129,7 @@ class PyTorchSklearnContainer(SklearnContainer):
             with open(os.path.join(location, constants.SAVE_LOAD_TORCH_JIT_PATH), "rb") as file:
                 container = dill.load(file)
         else:
+            shutil.rmtree(location)
             raise RuntimeError("Model type {} not recognized".format(model_type))
 
         # Need to set the number of threads to use as set in the original container.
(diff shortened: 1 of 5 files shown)


hummingbird/ml/containers/sklearn/pytorch_containers.py
```diff
@@ -80,6 +80,7 @@ def save(self, location):
             with open(os.path.join(location, constants.SAVE_LOAD_TORCH_JIT_PATH), "wb") as file:
                 dill.dump(self, file)
         else:
+            shutil.rmtree(location)
```
ksaur, line 83: Will this call ever fail and cause another exception? I'm just wondering if the directory fails to create for some reason (permissions maybe), and then we try to delete it, will there be some issue?

ksaur (commented):
Can we add some tests for these fail cases?

Also, what if the directory is already there, as in...HB didn't create the directory but the user happens to already have one by that name. I think that this doesn't solve that problem maybe.

interesaaat:
> Can we add some tests for these fail cases?
> 
> Also, what if the directory is already there, as in...HB didn't create the directory but the user happens to already have one by that name. I think that this doesn't solve that problem maybe.

In this case we will throw an error because the directory is already there. I can add a test for this. I can attach a timestamp to the location name otherwise.

hummingbird/ml/containers/sklearn/pytorch_containers.py
interesaaat: Good point! I think that if the directory creation fails, it will raise an exception on line 54. So at this point we should be safe to delete the dir here?

ksaur: I *think* that's right. But yah a test to double check would clear any doubts :)

ksaur (approved):

PR #504, opened 2021-04-24
Add modules versions when saving models

interesaaat:


hummingbird/ml/_utils.py
```diff
@@ -223,6 +224,34 @@ def load(location):
...
+    configurations = []
+    for module in args:
+        assert isinstance(module, ModuleType)
+        configurations.append("{}={}".format(str(module.__name__), str(module.__version__)))
+    return "\n".join(configurations)
+
+
+def check_dumped_versions(configurations, *args):
+    configurations = [configuration.strip() for configuration in configurations]
+    versions = {version.split("=")[0]: version.split("=")[1] for version in configurations}
```
ksaur, line 243: add some comments here for clarity

hummingbird/ml/containers/sklearn/onnx_containers.py
```diff
@@ -12,8 +12,10 @@
 import os
 import numpy as np
 import shutil
+import torch
 
-from hummingbird.ml._utils import onnx_runtime_installed, from_strings_to_ints
+import hummingbird
```
ksaur, line 18: I see you use this import of `hummingbird` on line 76 to get the versoin...is there additional overhead of always importing the whole thing here at the top vs only importing it if the user calls the `save` function?  (I'm not sure.)  I think it's fine either way

ksaur (approved):

hummingbird/ml/containers/sklearn/onnx_containers.py
interesaaat: Good point. I have no idea. I think that at this point `hummingbird` modules should be already imported, so it is ok?

PR #512, opened 2021-05-05
adding process id to unzipped folder incase multiple readers

ksaur:
When a user has multiple concurrent readers trying to unzip the model at the same time, it collides and weird things happen.

My first thought was to append the process ID; but maybe there is something better?


ksaur:
heh. worked on my machine. :-P

PR #519, opened 2021-06-03
Add prophet integration (with only trend prediction)

interesaaat:


hummingbird/ml/supported.py
```diff
@@ -339,6 +340,18 @@ def _build_onnxml_operator_list():
     return []
 
 
+def _build_prophet_operator_list():
+    """
+    List all suported XGBoost (Sklearn API) operators.
```
ksaur: change this comment i think

hummingbird/ml/supported.py
```diff
@@ -339,6 +340,18 @@ def _build_onnxml_operator_list():
...
+    """
+    if prophet_installed():
+        from prophet import Prophet
+
+        return [Prophet]
```
ksaur, line 350: I'm confused, is there only one?

hummingbird/ml/operator_converters/prophet.py
```diff
@@ -0,0 +1,83 @@
...
+        trend = trend * self.y_scale + self.floor
+        return trend
+
+
+def convert_prophet(operator, device=None, extra_config={}):
+    """
+    Converter for `prophet.Prohpet`
+
+    Args:
+        operator: An operator wrapping a `ai.onnx.ml.Imputer` model
```
ksaur: what is the 'operator' here?

hummingbird/ml/supported.py
interesaaat: yep.

hummingbird/ml/supported.py
interesaaat: Yea they only have one operator. (as far as I can tell)

hummingbird/ml/operator_converters/prophet.py
interesaaat: ops. Fixed.

ksaur:
It's not hitting your [unit-tests](https://app.codecov.io/gh/microsoft/hummingbird/compare/519/diff), i think you need to add "install prophet" to the cicd yaml

interesaaat:
> It's not hitting your [unit-tests](https://app.codecov.io/gh/microsoft/hummingbird/compare/519/diff), i think you need to add "install prophet" to the cicd yaml

Ah that is what was going on. Thanks for catching this! Sending a fix now.

ksaur:
Happy "the build fails" friday!!! :-D

```
=========================== short test summary info ============================
FAILED tests/test_backends.py::TestBackends::test_udf_torch - py4j.protocol.P...
FAILED tests/test_backends.py::TestBackends::test_udf_torch_jit_spark_file - ...
FAILED tests/test_sklearn_multioutput_regression.py::TestSklearnMultioutputRegressor::test_sklearn_regressor_chain
===== 3 failed, 486 passed, 112 skipped, 159 warnings in 218.92s (0:03:38) =====
```

I see this _might_ explain the first two diffing yours to the prior successful run:
```diff
<   Downloading pyspark-3.1.2.tar.gz (212.4 MB)
---
>   Downloading pyspark-3.1.1.tar.gz (212.3 MB)
```

The other one (`test_sklearn_regressor_chain`) is because we have torch 1.5.1 on the mac so these mismatches are expected; I didn't think this test should be hit.   Do we need to add ` @unittest.skipIf(LooseVersion(torch.__version__) < LooseVersion("1.6.0"),` to the [test](https://github.com/microsoft/hummingbird/blob/main/tests/test_sklearn_multioutput_regression.py#L44)?

ksaur:
Hmm actually the build is fine (see #520), so prophet must be bringing along some "new friends".

ksaur (approved):

ksaur:
Should we merge #523 first and make sure this works on the unpins?

interesaaat:
Yes, good idea!

PR #521, opened 2021-06-06
Added the function assert_torch_installed() for message consistency.

ghost:
Repeated messages is not a good practice as it can cause difficulty and inconsistency when somebody wants to change the message line. Consider adding another function that does the assertion job.

### DIFF
--- a/hummingbird/ml/convert.py
+++ b/hummingbird/ml/convert.py
@@ -15,7 +15,7 @@ from .operator_converters import constants
 from ._parse import parse_sklearn_api_model, parse_onnx_api_model, parse_sparkml_api_model
 from ._topology import convert as topology_converter
 from ._utils import (
-    torch_installed,
+    assert_torch_installed,
     lightgbm_installed,
     xgboost_installed,
     pandas_installed,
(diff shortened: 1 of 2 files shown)


hummingbird/ml/convert.py
```diff
@@ -15,7 +15,7 @@
 from ._parse import parse_sklearn_api_model, parse_onnx_api_model, parse_sparkml_api_model
 from ._topology import convert as topology_converter
 from ._utils import (
-    torch_installed,
+    assert_torch_installed,
```
ksaur, line 18: Can you please also change the other `_installed` (Ex lightgbm/xgeboot/pandas) to be prefixed with `assert` for consistency?

ksaur (changes requested):
Thanks, you are right, this is a better practice! :)

For consistency, can you please also change the others to match (see my comment below) and also throughout `convert.py` (ex `lightgbm_installed` on line 117)

hummingbird/ml/convert.py
ghost: @ksaur Done!

ghost: I've also fixed some spelling errors along the way. In this case, the word `XGBoost` and `install` was incorrectly spelled and capitalized!
image

ksaur (approved):
Thanks for the additional fixes! :)

I think there's just a whitespace lint thing, then we can merge it.

ghost:
@ksaur Oh, really sorry for that! It looks ok now. 🙂

Issue #495, opened 2021-04-14
XGBoost does not work for 1.4.0 for hummingbird


ksaur:
In #492 , we needed to pin to `xgboost<1.4.0` because `==1.4.0` ([released](https://github.com/dmlc/xgboost/commit/b9a4f3336af7c78f9f7b66f7c91a419108b7ec83) on April 12th 2021) broke our tests.  We need to investigate why.

From [the pipeline](https://github.com/microsoft/hummingbird/pull/490/checks?check_run_id=2328236386):

```
tests/test_xgboost_converter.py F..FF...sFF...FFF...sFF..            

=================================== FAILURES ===================================
__________ TestXGBoostConverter.test_float64_xgb_classifier_converter __________

self = <test_xgboost_converter.TestXGBoostConverter testMethod=test_float64_xgb_classifier_converter>

    @unittest.skipIf(not xgboost_installed(), reason="XGBoost test requires XGBoost installed")
    def test_float64_xgb_classifier_converter(self):
        warnings.filterwarnings("ignore")
        num_classes = 3
        for max_depth in [1, 3, 8, 10, 12]:
            model = xgb.XGBClassifier(n_estimators=10, max_depth=max_depth)
            np.random.seed(0)
            X = np.random.rand(100, 200)
            y = np.random.randint(num_classes, size=100)
    
            model.fit(X, y)
    
>           torch_model = hummingbird.ml.convert(model, "torch", [])

tests/test_xgboost_converter.py:175: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/hummingbird/ml/convert.py:433: in convert
    return _convert_common(model, backend, test_input, device, extra_config)
/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/hummingbird/ml/convert.py:383: in _convert_common
    return _convert_xgboost(model, backend_formatted, test_input, device, extra_config)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

model = XGBClassifier(base_score=0.5, booster='gbtree', colsample_bylevel=1,
[...]

Issue #454, opened 2021-03-05
Onnxruntime 1.7.0 breaks tests


ksaur:
https://github.com/microsoft/hummingbird/runs/2035682979

We need to investigate why `onnxruntime==1.7.0` breaks the tests. (Last known good at `1.6.0`)

ksaur:
Closed with #523 🎊

PR #523, opened 2021-06-14
Remove constraint on dependencies version

xadupre:
The setup requires to install too old versions of dependencies. This PR removes these constraints.


ksaur:
Hi @xadupre !  We pinned these to specific versions due to incompatibilities, and are trying to fix them as needed.  Are you blocked on a particular version of something?

ksaur:
See also #454

xadupre:
> Are you blocked on a particular version of something?

I'm refreshing onnxmltools CI which involves hummingbird-ml. More recent versions would be better. Right now I need to do `pip install hummingbird-ml --no-deps` to avoid CI to revert to older versions of scikit-learn, lightgbm. The errors I saw in this PR are very similar to what I had when updated sklearn-onnx for scikit-learn 0.24. I'll try to see if similar changes can fix the pipeline unless somebody is already working on it.

ksaur:
> I'll try to see if similar changes can fix the pipeline unless somebody is already working on it.

Thanks for you looking into this, no one is currently looking at it!  We would greatly appreciate your help!

The information I have about our known version issues is as follows:
* For  scikit-learn, there was some issue with gradient boosting, i think you might have just re-discovered :)
* For onnxruntime (#454), we haven't dug into much yet.
* For lightgbm, there was some issue (#389) I think related to onnxmltools.
* For XGB, there is an outstanding issue (#495) due to a breaking change in April. 

Please let me know if you have questions and we can try to sort things out on our end as well!

xadupre:
I may be able to fix all issues but this PR won't fix #389. I did a fix for this one in onnxmltools (https://github.com/onnx/onnxmltools/blob/master/onnxmltools/convert/lightgbm/operator_converters/LightGbm.py#L423). I'll see if I can easily do something similar.

xadupre:
scikit-learn and XGB are done.

tests/test_lightgbm_converter.py
```diff
@@ -225,8 +225,9 @@ def test_float64_lgbm_perf_tree_trav_regressor_converter(self):
         self._run_float64_lgbm_regressor_converter(1000, extra_config={"tree_implementation": "perf_tree_trav"})
 
     # Random forest in lgbm
+    @unittest.skipIf(True, reason="boosting_type=='rf' produces different probabilites.")
```
ksaur, line 235: Will this test ever be used?  Is this due to the runtime error added on lightgbm.py:69-72? If so, can you please add a comment?

hummingbird/ml/_parse.py
```diff
@@ -624,6 +624,8 @@ def _parse_onnx_single_operator(topology, operator):
 
     # Add the operator in the topology.
     alias = get_onnxml_api_operator_name(operator.op_type)
+    if alias is None:
+        alias = "ONNXML" + operator.op_type
```
ksaur, line 628: I am curious, in what cases does this happen?

ksaur (approved):
This is wonderful, thank you so much @xadupre!

hummingbird/ml/operator_converters/onnx/tree_ensemble.py
```diff
@@ -40,6 +40,9 @@ def _get_tree_infos_from_onnx_ml_operator(model):
             threshold = attr.floats
         elif attr.name == "class_weights" or attr.name == "target_weights":
             values = attr.floats
+            if len(values) == 0:
+                raise TypeError(
+                    "Type mismatch with attribute %r." % attr)
```
interesaaat: Nit: in HB we use `string.format()` instead of `%`. Cold you please use format for consistency?

interesaaat (approved):

hummingbird/ml/_parse.py
xadupre: StandardScaler is converted into an onnx Scaler when types are float. This operator only supports float, not double. So for double, the scaler is replaced by operator Sub, Mul. Alias is None is this case. The fix I did may be not the best one but it worked in this case.

hummingbird/ml/operator_converters/onnx/tree_ensemble.py
xadupre: done

tests/test_lightgbm_converter.py
xadupre: I let the test because I assume this case will be supported. I added a comment in that way. I can remove the test.

ksaur: a comment is perfect, thanks!

interesaaat:
Thanks @sdpython for the contribution!

PR #526, opened 2021-06-21
troubleshooting knn breaking

ksaur:
An update to scipy (1.7.0) breaks KNN in SKL for now.  Until they push a fix, let's pin to scipy 1.6.3 in the ci/cd .yml only.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ with open(README) as f:
 install_requires = [
     "numpy>=1.15",
     "onnxconverter-common>=1.6.0",
-    "scipy",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -39,7 +39,7 @@
     # The need each for these depends on which libraries you plan to convert from
     "xgboost>=0.90",
     "lightgbm>=2.2",
-    "prophet==1.0.1"
+    "prophet==1.0.1",
```
ksaur, line 42: somehow this was missed in prior lint

interesaaat (approved):

PR #532, opened 2021-07-14
Update LGBM references with scikit-learn in the sklearn_year_with_train.ipynb notebook

parulnith:
Corrected a typo where the `scikit-learn` models were mentioned as `LGBM` models.


ksaur (approved):

ksaur:
@interesaaat it looks like one of our onnx dependencies must have pushed out an update and broke the build...we'll have to take a look tomorrow :)

Issue #533, opened 2021-07-14
Tests are failing due to onnx changes


ksaur:
It seems there were several updates to [sklearn-onnx](https://github.com/onnx/sklearn-onnx/commits/master), and in [our workflow](https://github.com/microsoft/hummingbird/blob/main/.github/workflows/pythonapp.yml#L56) we pull from main.

Last known good run [here](https://github.com/microsoft/hummingbird/actions/runs/962178702).
Error run ( [here](https://github.com/microsoft/hummingbird/pull/532/checks?check_run_id=3063224415) ):

```python
=================================== FAILURES ===================================
___________________ TestBackends.test_onnx_zero_shape_input ____________________

self = <test_backends.TestBackends testMethod=test_onnx_zero_shape_input>

    @unittest.skipIf(
        not (onnx_ml_tools_installed() and onnx_runtime_installed()), reason="ONNXML test require ONNX, ORT and ONNXMLTOOLS"
    )
    def test_onnx_zero_shape_input(self):
        warnings.filterwarnings("ignore")
        max_depth = 10
        num_classes = 2
        if CastTransformer is None:
            model = GradientBoostingClassifier(n_estimators=10, max_depth=max_depth)
        else:
            # newer version of sklearn-onnx
            model = make_pipeline(
                CastTransformer(dtype=np.float32),
                GradientBoostingClassifier(n_estimators=10, max_depth=max_depth))
        np.random.seed(0)
        X = np.random.rand(100, 200)
        y = np.random.randint(num_classes, size=100)
    
        model.fit(X, y)
    
        # Create ONNX-ML model
>       onnx_ml_model = convert_sklearn(model, initial_types=[("input", DoubleTensorType([0, X.shape[1]]))], target_opset=11)

tests/test_backends.py:538: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/onnxmltools/convert/main.py:66: in convert_sklearn
    custom_conversion_functions, custom_shape_calculators)
/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/skl2onnx/convert.py:150: in convert_sklearn
    final_types=final_types)
[...]

PR #534, opened 2021-07-14
troubleshooting build failure

ksaur:
Troublehshooting #533


ksaur:
@interesaaat We can merge this now so we are unblocked (and can merge #532), and then get a better fix from @xadupre ?

interesaaat (approved):

PR #540, opened 2021-08-20
added support for lightgbm booster

### LINKED ISSUE #539: loading lightgbm.basic.Booster
Is there a way to directly convert a LightGBM booster object instead of using the sklearn API? 

With the sklearn API for lgb, I think you can only load file to a booster object, and I'm not sure how to initialize a lgb regressor from a booster object.

Currently when I try convert booster object directly I get: hummingbird.ml.exceptions.MissingConverter: Unable to find converter for model type <class 'lightgbm.basic.Booster'>.

marsupialtail:
Issue #539 : added support for LightGBM booster

### DIFF
--- a/hummingbird/ml/operator_converters/lightgbm.py
+++ b/hummingbird/ml/operator_converters/lightgbm.py
@@ -102,8 +102,20 @@ def convert_sklearn_lgbm_regressor(operator, device, extra_config):
 
     return convert_gbdt_common(operator, tree_infos, _get_tree_parameters, n_features, extra_config=extra_config)
 
+def convert_lgbm_booster(operator, device, extra_config):
+
+    assert operator is not None, "Cannot convert None operator"
+
+    # Get tree information out of the model.
+    n_features = len(operator.raw_operator.feature_name())
+    tree_infos = operator.raw_operator.dump_model()["tree_info"]
+
+    return convert_gbdt_common(operator, tree_infos, _get_tree_parameters, n_features, extra_config=extra_config)
+
 
 # Register the converters.
 register_converter("SklearnLGBMClassifier", convert_sklearn_lgbm_classifier)
 register_converter("SklearnLGBMRanker", convert_sklearn_lgbm_regressor)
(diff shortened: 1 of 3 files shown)


ksaur:
@marsupialtail Thanks so much for the PR! :) I think there are some lint things to check on in the CI/CD.  

Can you please also add tests inside of [test_lightgbm_converter.py](https://github.com/microsoft/hummingbird/blob/main/tests/test_lightgbm_converter.py) to make sure that all lines of `convert_lgbm_booster` hit coverage?  Thank you! And please let me know if you have questions; happy to help.

marsupialtail:
I added the tests based on the LGBRanker PR you referred. Not sure if that's what you are looking for.

ksaur:
@marsupialtail I think there is another lint thing.  (I tried to give permissions so the CI/CD would automatically run for you but it didn't seem to work.)  

PR looks great otherwise!  (cc: @interesaaat any comments? )

hummingbird/ml/operator_converters/lightgbm.py
```diff
@@ -102,8 +102,18 @@ def convert_sklearn_lgbm_regressor(operator, device, extra_config):
 
     return convert_gbdt_common(operator, tree_infos, _get_tree_parameters, n_features, extra_config=extra_config)
 
+def convert_lgbm_booster(operator, device, extra_config):
+
```
ksaur, line 114: Can you please add the docstring info? (see lines 84-94 in this same file)

ksaur (changes requested):

ksaur:
Ok I *think* I fixed it so you should be able to run the ci/cd without additional permissions. (Github must have changed some setting, because the subsequent run should have 'just ran')

There's some formatting issue, please take a look. Thanks!

marsupialtail:
Sorry but I absolutely hate lint. Not sure if I will have too much motivation to finish fixing the lint problems :-(

ksaur (approved):

ksaur:
Thanks for the contribution and for sticking with it @marsupialtail !

PR #542, opened 2021-09-21
Fix few bugs in benchmarks

interesaaat:
New versions of `xgboost` and `openml` creates problems in the tree and pipelines benchmarks. This PR fixes this

### DIFF
--- a/benchmarks/trees/score.py
+++ b/benchmarks/trees/score.py
@@ -154,8 +154,9 @@ class ONNXMLBackend(ScoreBackend):
         with Timer() as t:
             if self.params["operator"] == "xgb":
                 initial_type = [("input", FloatTensorType([1, self.params["input_size"]]))]
-                fixed_names = list(map(lambda x: str(x), range(len(model._Booster.feature_names))))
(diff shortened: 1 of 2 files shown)


benchmarks/trees/score.py
```diff
@@ -207,7 +208,10 @@ def predict(self, predict_data):
                 end = min(start + batch_size, total_size)
 
                 if self.params["operator"] == "xgb":
-                    self.predictions[start:end, :] = self.sess.run([output_name], {input_name: predict_data[start:end, :]})
+                    pred = self.sess.run([output_name], {input_name: predict_data[start:end, :]})
+                    if type(pred) is list:
```
ksaur, line 212: is `pred` ever an empty list?  (like if error?)

ksaur (approved):

benchmarks/trees/score.py
interesaaat: If there is an error this should crash instead of returning an empty list. If the input is empty probably we will get an empty list tho, but this should not happen.

PR #560, opened 2021-12-16
cleaning up versions for python3.9

### LINKED ISSUE #555: Python 3.6 EOL
Python 3.6 is [EOL on Dec 23, 2021.](https://endoflife.date/python).  

We need to remove that and add python 3.9 (and possibly 3.10) to the ci/cd pipeline.
This also means ditching pytorch < 1.6. support.

ksaur:


interesaaat (approved):

README.md
```diff
@@ -65,9 +64,9 @@ _Thank you to [Chien Vu](https://www.linkedin.com/in/vumichien/) for contributin
 
 ## Installation
 
-Hummingbird was tested on Python >= 3.6 on Linux, Windows and MacOS machines.  (Python 3.5 is suppored up to `hummingbird-ml==0.2.1`.)  It is recommended to use a virtual environment (See: [python3 venv doc](https://docs.python.org/3/tutorial/venv.html) or [Using Python environments in VS Code](https://code.visualstudio.com/docs/python/environments).)
+Hummingbird was tested on Python >= 3.7 on Linux, Windows and MacOS machines.  (Python 3.6 is suppored up to `hummingbird-ml==0.4.2`.)  It is recommended to use a virtual environment (See: [python3 venv doc](https://docs.python.org/3/tutorial/venv.html) or [Using Python environments in VS Code](https://code.visualstudio.com/docs/python/environments).)
```
interesaaat: Ah typo! suppored -> supported

hummingbird/ml/_topology.py
```diff
@@ -208,11 +210,6 @@ def convert(topology, backend, test_input, device, extra_config={}):
             )
 
         if backend == onnx.__name__:
-            # vers = LooseVersion(torch.__version__)
-            # allowed_min = LooseVersion("1.6.0")
-            # Pytorch <= 1.6.0 has a bug with exporting GEMM into ONNX.
-            # For the moment only tree_trav is enabled for pytorch <= 1.6.0
-            # if vers < allowed_min:
             extra_config[constants.TREE_IMPLEMENTATION] = "tree_trav"
```
interesaaat: Can we please leave this comment. Eventually when we will move to PyTorch > 1.6 we could remove the condition (assuming they fixed the bug)

tests/test_sklearn_linear_converter.py
```diff
@@ -202,7 +202,7 @@ def test_sgd_classifier_multi(self):
 
     # SGDClassifier with modified huber loss
     @unittest.skipIf(
-        LooseVersion(torch.__version__) < LooseVersion("1.6.0"), reason="Modified Huber loss test requires torch >= 1.6.0"
```
interesaaat, line 205: Wait can we just remove this? Now we only support Pytorch >= 1.6.0.

tests/test_sklearn_linear_converter.py
```diff
@@ -218,7 +218,7 @@ def test_modified_huber(self):
         np.testing.assert_allclose(model.predict_proba(inputs), hb_model.predict_proba(inputs), rtol=1e-6, atol=1e-6)
 
     @unittest.skipIf(
-        LooseVersion(torch.__version__) < LooseVersion("1.6.0"), reason="Modified Huber loss test requires torch >= 1.6.0"
+        LooseVersion(torch.__version__) < LooseVersion("1.7.0"), reason="Modified Huber loss test requires torch >= 1.7.0"
```
interesaaat, line 221: Same here.

tests/test_sklearn_linear_converter.py
```diff
@@ -234,7 +234,7 @@ def test_modified_huber2(self):
 
     # SGDClassifier with modified huber loss multiclass
     @unittest.skipIf(
-        LooseVersion(torch.__version__) < LooseVersion("1.6.0"), reason="Modified Huber loss test requires torch >= 1.6.0"
+        LooseVersion(torch.__version__) < LooseVersion("1.7.0"), reason="Modified Huber loss test requires torch >= 1.7.0"
```
interesaaat, line 237: Same here.

.github/workflows/pythonapp.yml
```diff
@@ -36,19 +36,19 @@ jobs:
         # The GitHub editor is 127 chars wide
         flake8 . --count  --max-complexity=10 --max-line-length=127 --statistics
     # PyTorch for Mac has different pip syntax wrt Win and Linux.
-    # We want to test torch 1.5.x (in this case on python3.6) and torch 1.7.x.
+    # We want to test torch 1.6.0 (in this case on python3.7) and torch 1.10.x.
```
interesaaat, line 39: Why do we do this? I don't remember. Is this because for 3.5 we were only able to run PyTorch 1.5.1? If true maybe we can just test on 1.10.x and remove all these conditions.

hummingbird/ml/_topology.py
ksaur: Ok, to clarify you mean leave the comment but still remove the condition?

tests/test_sklearn_linear_converter.py
ksaur: This one is ok to remove because we are only supporting >=1.6, so it doesn't apply to < 1.6.

tests/test_sklearn_linear_converter.py
ksaur: In one of my commit messages, i have something like "turns out, clip is in torch 1.7 not 1.6, so adding back loose versio and putting it to the right number".

Modified Huber Loss is not in 1.6 apparently.

.github/workflows/pythonapp.yml
ksaur: Now we are testing python 3.7 on torch 1.6, and 1.10 on all other pythons.

I think we did it just to test older torch versions?

You are right, we could just 1.10 on ALL of these...  But I guess that helped us test our `LooseVersion` bounds, so maybe we leave it?

tests/test_sklearn_linear_converter.py
interesaaat: Wait but it is working now with torch 1.6, no?

hummingbird/ml/_topology.py
interesaaat: I think we can leave this as is right now. In the future we may want to remove the condition (as well as the comment). Like if the minimum version of pytorch we support will become 1.7 and onnx works.

tests/test_sklearn_linear_converter.py
ksaur: No, it's not working with torch 1.6. The test failed, so I looked into it, and that is why I changed it to this.

Issue #551, opened 2021-12-02
Error when converting Json model generated by XGBoost to Pytorch 


rodsang:
Hello,

I am trying to convert my model into a pytorch model. My model is generated by XGBoost in python and is saved in the Json format, as follows:
.
.
.

X_train, X_test, y_train, y_test = train_test_split(X, Y, test_size=0.2, stratify=Y, random_state=1)
enc = OneHotEncoder(handle_unknown='ignore')
Y_train = enc.fit_transform(y_train.values.reshape(-1,1)).toarray()
Y_test = enc.fit_transform(y_test.values.reshape(-1,1)).toarray()
train = np.argmax(Y_train,axis =  1)
test = np.argmax(Y_test, axis = 1)
clf = XGBClassifier(learning_rate=0.1,
                    n_estimators=400,
                    max_depth=5,
                    min_child_weight=1,
                    gamma=0,
                    subsample=0.8,
                    colsample_bytree=0.8,
                    objective='multi:softprob',
                    nthread=4,
                    num_class=3,
                    seed=27,
                    predictor = 'gpu_predictor',
                    tree_method='gpu_hist',
                    gpu_id=0)

clf.fit(X_train,train.ravel(),
            verbose=False,
            early_stopping_rounds=50,
            eval_metric='merror',
            eval_set=[(X_test, test.ravel())])

y_predict = clf.predict(X_test)
y_train_predict = clf.predict(X_train)

clf.save_model('model.json')

model = XGBClassifier()

model.load_model('model.json')

# here i try to convert it

from hummingbird.ml import convert

modelc = convert(model, 'pytorch', extra_config={"n_features":54})
modelc.save('hb_model')

###ERROR

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-157-747a9a976fe6> in <module>
      4 
      5 
----> 6 modelc = convert(model, 'pytorch', extra_config={"n_features":54})
      7 #modelc.save('hb_model')

.
.
.

ValueError: invalid literal for int() with base 10: 'chroma_stft
:::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::

Do you know want can i possibly be doing wrong?

interesaaat:
Hi! Maybe but converting the model to JSON and back some field get erased
in the process. I will look into it and keep you posted.

rodsang:
> 

[...]

PR #562, opened 2022-01-08
Fix problem with pandas used with latest xgb models

### LINKED ISSUE #561: incompatible with XGBoost 1.5.0 due to breaking model format changes
In xgboost 1.5.0 release, a breaking change is introduced to the model format causing this error when trying to use `convert`:
```
ValueError: invalid literal for int() with base 10: '<some_feature_name>'
```
The error happens because this line: https://github.com/microsoft/hummingbird/blob/main/hummingbird/ml/operator_converters/xgb.py#L33

Previously, XGBoost seemed to use an indexed integer to store the feature name, but now it's changed to the alphabet name of the feature. I tried changing the conversion to use other bases, say `36` to get some 1-1 integer map to the feature name, but is causing another issue down the road in the indexing. 
Seemed like the lib may need to build some indexing to the feature.

Dataset used: `sklearn.datasets.load_boston`

interesaaat:
Closes #561 and #551.

### DIFF
--- a/hummingbird/ml/operator_converters/xgb.py
+++ b/hummingbird/ml/operator_converters/xgb.py
@@ -59,7 +59,7 @@ def _tree_traversal(tree_info, lefts, rights, features, thresholds, values):
             count += 1
 
 
-def _get_tree_parameters(tree_info):
+def _get_tree_parameters(tree_info, extra_config):
     """
     Parse the tree and returns an in-memory friendly representation of its structure.
     """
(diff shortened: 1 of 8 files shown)


ksaur (approved):

hummingbird/ml/operator_converters/xgb.py
```diff
@@ -68,6 +68,12 @@ def _get_tree_parameters(tree_info):
     features = []
     thresholds = []
     values = []
+
+    # Subsitute feature names with ids if necessary.
+    if constants.FEATURE_NAMES in extra_config:
```
ksaur, line 73: nice fix. :)

PR #566, opened 2022-02-07
allow tree_implementation="gemm" with onnx backend

### LINKED ISSUE #564: tree_implementation parameter gets overwritten with onnx backend
Hello, 

I have been trying to use the gemm with onnx backend but the onnx graph did not really look like we were using gemm tree implementation:
```python
model = convert(skl_model, backend="onnx", test_input=X, extra_config={"tree_implementation":"gemm"})
```
It seems that the `tree_implementation` parameter gets overwritten deeper in the code:

```python
        if backend == onnx.__name__:
            # vers = LooseVersion(torch.__version__)
            # allowed_min = LooseVersion("1.6.0")
            # Pytorch <= 1.6.0 has a bug with exporting GEMM into ONNX.
            # For the moment only tree_trav is enabled for pytorch <= 1.6.0
            # if vers < allowed_min:
            extra_config[constants.TREE_IMPLEMENTATION] = "tree_trav"
```
I am not sure why this is done tho? Removing this part gives me back the expected behavior.

Could you give a bit more info on what is the point of this? Can I do anything to fix this?

jfrery:
closes #564 

Allows onnx backend with TREE_IMPLEMENTATION="gemm".

### DIFF
--- a/hummingbird/ml/_topology.py
+++ b/hummingbird/ml/_topology.py
@@ -9,6 +9,7 @@ Converters for topology IR are stored in this file.
 """
 import numpy as np
 import os
+from distutils.version import LooseVersion
 import torch
 from uuid import uuid4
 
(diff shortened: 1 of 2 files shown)


interesaaat:
It looks that the exception here is not related to your PR. I am looking into it (they may have updated the metrics in the latest sklearn release).

ksaur:
Hi @jfrery Thanks for the PR!! Can you please rebase with Matteo's fix and try again? Pipeline should be working now. :)

interesaaat (approved):

hummingbird/ml/_topology.py
```diff
@@ -9,6 +9,7 @@
 """
 import numpy as np
 import os
+from packaging import version
```
interesaaat: A super minor thing. Throughout Hummingbird we use `from distutils.version import LooseVersion` (instead of this) and then we do for example `LooseVersion(torch.__version__) <= LooseVersion("1.4")`. Do you think you can use the same approach? Just for consistency really.

interesaaat (changes requested):
A super minor thing

hummingbird/ml/_topology.py
jfrery: Of course. Done.

interesaaat (approved):

PR #576, opened 2022-04-25
Fix xgboost tests for xgb > 1.6.0

interesaaat:
With the newer version of XGBoost the test creating a shallow tree with 1 node fail. This PR fixes the tests.


ksaur (approved):
Thanks for the quick fix!

PR #583, opened 2022-06-05
Changes in README.md file of hummingbird

bishtanuj:
Replace the images made by hands with the digital ones.
Hope you like it!


ksaur:
Thanks for the thought, but the images we had weren't made by hand, they were custom styled that way and we picked those images because we really liked the artwork.

bishtanuj:
> Thanks for the thought, but the images we had weren't made by hand, they were custom styled that way and we picked those images because we really liked the artwork.

Okay @ksaur
Sorry for changing the images.

PR #586, opened 2022-06-13
Add Python 3.10

mshr-h:


mshr-h:
ONNX has supported Python 3.10 after ONNX 1.11 but they haven't released an official Python 3.10 wheel. In the next release, there will be Python 3.10 wheel. ONNX 1.12 will be released in 2-3 weeks according to https://github.com/onnx/onnx/issues/4250.

ksaur:
I remember a similar delay on the ONNX'es when I ported us to Python3.9. Thanks for looking into 3.10

mshr-h:
ONNX Runtime 1.12 will be released mid-July with Python 3.10 support.
https://github.com/microsoft/onnxruntime/projects/9

ksaur:
Hi @mshr-h, I'm guessing the failure was because ONNX for py3.10 isn't out yet?

Also, I just realized...we changed the pipeline to be `macos==11` instead of `macos==10.15`.  Is that also [missing](https://pypi.org/project/onnxruntime/#files) from onnxruntime?  I'm a little surprised since macos11 came out in 2020...

ksaur:
Ah, nevermind, I see [here](https://github.com/microsoft/hummingbird/runs/6976226331?check_suite_focus=true) that it appears the macos11 run just used the onnxruntime  for mac 10.14, so maybe it's ok!

mshr-h:
@ksaur @interesaaat 
Can you review it?

ksaur (dismissed):
Thanks for all your work on this! Can you please also change the README.md, it mentions the versions in the badge and the text ([example of 3.9 change](https://github.com/microsoft/hummingbird/pull/560/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5))? Thanks!

interesaaat:
I think something is wrong with the TVM cache.

interesaaat (approved):

PR #588, opened 2022-06-15
Fix the things broken by SKL==1.1.1

### LINKED ISSUE #584: New scikit learn release breaks SklearnHistGradientBoostingClassifier tests
In the [pipeline](https://github.com/microsoft/hummingbird/runs/6732643088?check_suite_focus=true) we hit an issue with `scikit-learn==1.1.1`.  These are all related to a shape mismatch in `HistGradientBoostingClassifier`.

example:
```python
self = <hummingbird.ml.operator_converters._tree_commons.ApplyBasePredictionPostTransform object at 0x7f787fb3af40>
x = tensor([[ 0.1304],
        [-0.1389],
        [-0.0951],
        [ 0.1901],
        [ 0.0985],
        [-0.0525],
    ...50],
        [-0.1293],
        [-0.2084],
        [-0.2635],
        [ 0.3453],
        [ 0.0253],
        [ 0.1689]])
    def __call__(self, x):
>       x += self.base_prediction
E       RuntimeError: output with shape [100, 1] doesn't match the broadcast shape [1, 1, 100, 1]
```
We get this same error on 10 of our tests (9 in `test_sklearn_histgbdt_converters.py` and 1 in `test_sklearn_decomposition.py`.  For now (in #582) we pinned `"scikit-learn>=0.21.3,<1.1.0"`.

It seems that  in v1.1.1 [there were many changes, including an API change](https://scikit-learn.org/stable/whats_new/v1.1.html), to `HistGradientBoostingClassifier`.  I spent an hour looking at it, but nothing obvious jumped out.

### LINKED ISSUE #587: output with shape [369, 1] doesn't match the broadcast shape [1, 1, 369, 1]
I have a sklearn's HistGradientBoostingRegressor trained on [369, 80] data to output [369, 1] result. After converting it to Hummingbird model, I'm getting this error on the very same data.

ksaur:
Looking forward to breaking additional things in the process! :)


ksaur:
closes #584 and #587

ksaur:
This "fixes" HistGB, but new other [error](https://github.com/microsoft/hummingbird/runs/6892676674?check_suite_focus=true) on `test_pca_converter_two_no_scale`:

```
        comparison_operator = (
            operator.gt if include_boundaries in ("right", "both") else operator.ge
        )
        if max_val is not None and comparison_operator(x, max_val):
>           raise ValueError(
                f"{name} == {x}, must be"
                f" {'<=' if include_boundaries in ('right', 'both') else '<'} {max_val}."
            )
E           ValueError: n_components == 10, must be <= 3.
```

hummingbird/ml/operator_converters/_gbdt_commons.py
```diff
@@ -134,7 +134,14 @@ def convert_gbdt_common(
 
     # Define the post transform.
     if constants.BASE_PREDICTION in extra_config:
-        base_prediction = torch.nn.Parameter(torch.FloatTensor(extra_config[constants.BASE_PREDICTION]), requires_grad=False)
+
+        base_pred = torch.FloatTensor(extra_config[constants.BASE_PREDICTION])
+
+        # For newer versions of scikit-learn (>1.1.1),
+        if len(base_pred.shape) == 4:
+            base_pred = torch.nn.Parameter(torch.FloatTensor(base_pred[0][0]), requires_grad=False)
```
ksaur: I'm open to suggestions; there is likely a better way

interesaaat: I think this is good. No idea why they get this shape.

interesaaat: No need to do `torch.nn.Parameter(torch.FloatTensor(`, I think that just `base_pred[0][0]` is enough.

tests/test_sklearn_decomposition.py
```diff
@@ -159,7 +159,7 @@ def test_pca_converter_two(self):
 
     # PLS regressor n_componenets two no scale
     def test_pca_converter_two_no_scale(self):
-        self._fit_model_pls_regressor(PLSR(n_components=10, scale=False))
+        self._fit_model_pls_regressor(PLSR(n_components=3, scale=False))
```
interesaaat, line 162: Does this change anything?

interesaaat (changes requested):

tests/test_sklearn_decomposition.py
ksaur: Yes, with the skl==1.1.1, you'll get [ValueError: n_components == 10, must be <= 3.](https://github.com/microsoft/hummingbird/runs/6892676674?check_suite_focus=true)

hummingbird/ml/operator_converters/_gbdt_commons.py
ksaur: ah you're right, thanks, fixed

interesaaat (approved):

interesaaat:
Thanks Karla for fixing this!

PR #592, opened 2022-06-15
Use `onnxmltools>=1.6.0,<=1.11.0` instead of `onnxmltools>=1.6.0`

### LINKED ISSUE #589: Code coverage
Code coverage dropped to 80% from 90% unexpectedly. TODO: investigate

mshr-h:
`import onnxmltools` will fail with
`onnxconverter_common==1.8.1` & `onnxmltools==1.11.1`. The failure caused ONNX test skipping. 
In this PR, we'll use `onnxmltools>=1.6.0,<=1.11.0` instead of `onnxmltools>=1.6.0`.

Fix #589

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -74,6 +74,8 @@ jobs:
       run: |
         pip install .[extra,onnx,sparkml]
         pip install pandas
+    - name: pip list
+      run: pip list
     - uses: actions/cache@v1
       # TVM takes forever, we try to cache it.
       if: ${{ matrix.os != 'windows-2019' }}
@@ -126,6 +128,8 @@ jobs:
       working-directory: ../../../tvm/python
       run: |
         python setup.py install
+    - name: pip list
+      run: pip list
     # We don't run pytest for Linux py3.7 since we do coverage for that case.
     - name: Test with pytest
       if: ${{ matrix.python-version != '3.7' || matrix.os != 'ubuntu-18.04' }}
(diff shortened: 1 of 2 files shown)


ksaur:
This is great, thank you for finding the issue!!   Not even related to your PR/TVM,  but just unlucky timing.

This  work for me ([back to 90%](https://app.codecov.io/gh/microsoft/hummingbird/compare/592?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=microsoft)), what do you think @interesaaat ?

interesaaat:
This is great, thanks for finding this. Let's merge it.

ksaur:
We had to pin it to that exact version of skltoonnx because of #533.   :(

I see that when you merged it, it [passed](https://github.com/microsoft/hummingbird/runs/6910502198?check_suite_focus=true) on python 3.7 but failed on python 3.8?

I'm confused how it passed before... was it bumping to skl==1.1.1 that broke things for you?  Meaning, if  you set everything back to [f452692](https://github.com/microsoft/hummingbird/pull/592/commits/f452692e92e86b8bd9c22c570b4ef24abc67723e), is it ok then?

mshr-h:
On Python 3.7, `pip install git+https://github.com/onnx/sklearn-onnx.git@9b551d8ecfd9f97ad8771909e34cfffea7b6e99d` installs scikit-learn-1.0.2. But on Python 3.8, it installs scikit-learn-1.1.1.
I don't know why...

mshr-h:
I think bumping to skl==1.1.1 broke something. With skl==1.0.2, CI has passed. Should we revert [2fe2e35](https://github.com/microsoft/hummingbird/commit/2fe2e35e2ba113bc87cbebd3508271f922032d61)?
What do you think? @ksaur @interesaaat

mshr-h:
Another option is that we first go back to skl=1.0.2, then merge this commit. After that bump to skl=1.1.0.

ksaur (approved):

ksaur:
Thanks so much for all the time you put into this; I really appreciate it!

I don't want to revert the last commit entirely because it had fixes we'll need moving forward, but we can definitely go back to skl1.0.2 for now.

PR #593, opened 2022-06-16
Updating library versions

ksaur:
"scikit-learn requires: Python (>= 3.8)" [link](https://pypi.org/project/scikit-learn/)

Deprecating Python 3.7.  Removing the skl2onnx pin.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -53,7 +53,7 @@ jobs:
     - name: Install basic dependencies
       run: |
         pip install .[tests] -f https://download.pytorch.org/whl/torch_stable.html
-        pip install "scikit-learn>1.1.0"
+        pip install scikit-learn
     - name: Run basic tests without extra
       run: pytest
     - name: Coverage on basic tests without extra
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -54,7 +54,7 @@ jobs:
       run: |
         pip install .[tests] -f https://download.pytorch.org/whl/torch_stable.html
         pip install git+https://github.com/onnx/sklearn-onnx.git@9b551d8ecfd9f97ad8771909e34cfffea7b6e99d
-        pip install "scikit-learn>=0.21.3,<1.1.0"
+        pip install "scikit-learn>1.1.0"
     - name: Run basic tests without extra
       run: pytest
     - name: Coverage on basic tests without extra
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, macos-10.15, windows-2019]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


ksaur:
Ok, it fails on the skl2onnx with `log_loss` on 5 tests:

```
=========================== short test summary info ============================
FAILED tests/test_backends.py::TestBackends::test_onnx_no_test_data_double - ...
FAILED tests/test_backends.py::TestBackends::test_onnx_no_test_data_float - N...
FAILED tests/test_backends.py::TestBackends::test_onnx_zero_shape_input - Not...
FAILED tests/test_extra_conf.py::TestExtraConf::test_onnx_deafault_n_threads
FAILED tests/test_extra_conf.py::TestExtraConf::test_pandas_batch_onnxml - No...
===== 5 failed, 595 passed, 11 skipped, 380 warnings in 611.54s (0:10:11) ======
```

Need to check what the errors are if we remove the skl2onnx pin.

.github/workflows/pythonapp.yml
```diff
@@ -53,8 +53,7 @@ jobs:
     - name: Install basic dependencies
       run: |
         pip install .[tests] -f https://download.pytorch.org/whl/torch_stable.html
-        pip install git+https://github.com/onnx/sklearn-onnx.git@9b551d8ecfd9f97ad8771909e34cfffea7b6e99d
-        pip install "scikit-learn>=0.21.3,<1.1.0"
+        pip install scikit-learn
```
mshr-h: I think it's better to put `scikit-learn` to the `setup.py` instead of installing `scikit-learn` explicitly.

ksaur: Haha yes, thanks, I was putting it here to troubleshoot in this draft. But you're right, it seems it was removed by mistake in a prior PR.  I'll fix it after I am done troubleshooting.

ksaur:
With the new skl2onnx version, the StandardScaler with mean/std false

```python
model = StandardScaler(with_mean=False, with_std=False)
onnx_ml_model = convert_sklearn(model, initial_types=[("float_input", FloatTensorType([None, X.shape[1]]))])
```
returns the following graph:
```
[input: "float_input"
output: "variable"
name: "N0"
op_type: "Identity"
domain: ""
]
```
Note that there are no operators here, just `identity`, which is why  `test_standard_scaler_onnx_ff` fails.  (There are no operators because we [skip identity](https://github.com/microsoft/hummingbird/blob/3ce519a7ae26da67b66b54c263412ff7e1e6b315/hummingbird/ml/_parse.py#L622), so when it tries to parse inputs/outputs, it encounters several errors.)  It's not clear to me if this is expected behavior or a bug in `convert_sklearn`.

For example ` model = StandardScaler(with_mean=True, with_std=False)` returns:

```
[input: "float_input"
output: "variable"
name: "Scaler"
op_type: "Scaler"
attribute {
...
```

I'm going to throw an error here instead, as it seems not supported. The question is _where_ to throw.

ksaur:
The other failing onnx test `test_backends.py:test_onnx_zero_shape_input` is failing on the onnx convert before it touches Hummingbird:

```python
onnx_ml_model = convert_sklearn(model, initial_types=[("input", DoubleTensorType([0, X.shape[1]]))], target_opset=11)

RuntimeError: A variable cannot be empty, raw_name='input', onnx_name='input', shape=[0, 200], type=DoubleTensorType(shape=[0, 200]).
```

Since they are disallowing that we create this, i will delete this test.

ksaur:
After this, we need to have a discussion about deprecating torch==1.7.  I left this (test old toch) in the pipeline logic for now in case we want to have something test for it, but I think after this I'll do another PR removing `if torch==1.7` stuff.

interesaaat (approved):

PR #594, opened 2022-06-17
deprecate torch1.7, push to macOS11

ksaur:
For after #593, deprecate torch==1.7, as it is 2+ years old. (Current  version is torch==1.11.0).

Update mac in pipeline to 11.

TODO: there seems to be a bug with sphinx and newer torches, leaving that one as-is for now.


ksaur:
TODO: merge the other one first, then update the build pipeline gates w/new macos in the branch settings

interesaaat (approved):

PR #605, opened 2022-07-05
Use `python -m pip` instead of the `pip` executable

mshr-h:
Because it's the preferred invocation method.

[Issue 22295: Clarify available commands for package installation - Python tracker](https://bugs.python.org/issue22295)


ksaur:
LGTM! @interesaaat can you double check? :)

hummingbird/ml/operator_converters/sklearn/gbdt.py
```diff
@@ -18,7 +18,7 @@
 
 def _get_n_features(model):
     try:
-        return model.n_features_
+        return model.n_features_in_
     except AttributeError:
         # HistGradientBoosting
         return model._n_features
```
interesaaat: For HistGradientBoosting should we also use `n_features_in_`?

mshr-h: In the official docs, there's `n_features_in_` in `HistGradientBoosting` but not `n_features_`.
I think we should use `n_features_in_` for `HistGradientBoosting`.

[sklearn.ensemble.HistGradientBoostingClassifier — scikit-learn 1.1.1 documentation](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingClassifier.html)
[sklearn.ensemble.HistGradientBoostingRegressor — scikit-learn 1.1.1 documentation](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingRegressor.html)

interesaaat: Ok, let's remove the `except` part and use `n_features_in_` everywhere?

mshr-h: Sounds good. To keep the PR small, we should work on another branch.
I've opened a new issue #606 to track that.

interesaaat (approved):

PR #608, opened 2022-07-28
Upgrade prophet to v1.1

mshr-h:
`prophet==1.0.1` depends on `pystan==2.19.1.1` which is no longer maintained and not compatible with Python 3.10. They removed the dependency in the [`prophet==1.1` release](https://github.com/facebook/prophet/releases/tag/v1.1).
In this PR, we bump to `prophet==1.1`.
Related to #586


mshr-h:
@ksaur @interesaaat 
Can you please review it? Thanks.

ksaur (approved):

Issue #613, opened 2022-08-03
Make onnx dependencies optional


jppgks:
hummingbird-ml depends on onnxconverter-common. onnxconverter-common 1.12 introduced a dependency on onnxruntime. We had to pin onnxconverter-common: https://github.com/ludwig-ai/ludwig/pull/2347.

At Ludwig, we do not use the onnx features from hummingbird, so we'd like to avoid having to install them.

ksaur:
We use `onnxconverter-common` in our internal IR.  I'm not sure why onnxconverter-common 1.12 introduced a dependency on onnxruntime  (https://github.com/microsoft/onnxruntime/issues/11640).   

@xadupre Do you have any suggestions?

As a temporary fix, since we already have `"onnxmltools>=1.6.0,<=1.11.0"` in the [optional deps](https://github.com/microsoft/hummingbird/blob/main/setup.py#L36) due to #592, we could also make it `"onnxconverter-common>=1.6.0,<=1.11.0"` to avoid inadvertantly taking a dependency on onnxruntime

xadupre:
I'll look into that. This is a surprise to me. There is no mention of onnxruntime in onnxconverter-common on github.

mshr-h:
HB 0.4.4 depends on `onnxmltools>=1.6.0`. The `onnxmltools==1.12.0` tries to import [onnxconverter-common.auto_mixed_precision.py](https://github.com/microsoft/onnxconverter-common/blob/master/onnxconverter_common/auto_mixed_precision.py) which [depends on onnxruntime](https://github.com/microsoft/onnxconverter-common/blob/master/onnxconverter_common/auto_mixed_precision.py#L29).
I think that's why ...

As mentioned by @ksaur, we've temporarily fixed the issue with #592. But we haven't released the stable release with it.

@ksaur @interesaaat 
I think it's a good time to release v0.4.5. What do you think?

xiaowuhu:
update onnxconverter-common to 1.12.1 to fix this issue, please have a try.

PR #614, opened 2022-08-03
Pinning onnxconverter-common to avoid dep

ksaur:
Temp fix for #613


interesaaat (approved):

ksaur:
addresses https://github.com/ludwig-ai/ludwig/pull/2347 for now!

Issue #611, opened 2022-08-02
[CI] TVM cache is broken


mshr-h:
https://github.com/microsoft/hummingbird/issues/586#issuecomment-1201529514_

Link to the CI result page: [Add Python 3.10 · microsoft/hummingbird@3c52b20](https://github.com/microsoft/hummingbird/runs/7615498689?check_suite_focus=true)

TODO: investigate

ideas worth trying
- [ ] execute build even if TVM cache is hit
- [ ] try to create pip package and only cache it (do not cache the entire TVM directory)
- [ ] try ccache
- [x] update `actions/cache@v1` to `actions/cache@v3`

ksaur:
This seems to happen about once every 4-8 weeks.  I'm not sure if it's something we can control (on github's end?  I think?).  It's definitely been going on for a while  ([example](https://github.com/microsoft/hummingbird/issues/506))!

interesaaat:
Still failing :(

ksaur:
Right, the time we did get the green checkmark, it was because it rebuilt from scratch (which of course took 51 minutes...)

This is a new level of broken I think! 🤣

ksaur:
Apparently we can somehow [clear it](https://docs.github.com/en/rest/actions/cache#delete-github-actions-caches-for-a-repository-using-a-cache-key) with the API

ksaur:
One weird thing...when I run the github API (to delete the MacOS TVM caches so we can move forward temporarily by rebuilding), I only ever see the MacOS ones saved.

I see the following  cache keys:
`key": "macOS-7-3.9-tvm-0.9", "key": "macOS-7-3.10-tvm-0.9","key": "macOS-7-3.8-tvm-0.9","key": "macOS-6-3.9-tvm-0.9","key": "macOS-6-3.10-tvm-0.9","key": "macOS-6-3.8-tvm-0.9",`

It looks like #609 changed line 85 to cache mac only instead of "not windows", which means  Linux is never cached, unless I am misunderstanding something?   That probably isn't related to our current issue, but maybe there is something else going on related to it?

ksaur:
Ah ok, nevermind. i see that on [line 80](https://github.com/microsoft/hummingbird/blob/main/.github/workflows/pythonapp.yml#L80) we always pip install Linux TVM now.

I was just trying to think back to a time when the cache worked more consistently and if something unexpected changed!

PR #615, opened 2022-08-04
Use pre-installed LLVM for building TVM on macOS-11

mshr-h:
Related to #611 
GitHub-hosted runner has Clang/LLVM by default.
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md


mshr-h:
I’m not confident this will solve the TVM build error but it’s definitely better to use pre-installed LLVM package than manually installing it.

@ksaur @interesaaat 
Can you review it?
Thanks!

ksaur (approved):
Thanks for looking into this!

PR #617, opened 2022-08-04
bump to v0.4.5

ksaur:


ksaur:
We're hitting a bug with TVM install on macos11 (maybe we wait for #615)

interesaaat (approved):

PR #618, opened 2022-08-08
Remove pinned version for onnxconverter-common

### LINKED ISSUE #613: Make onnx dependencies optional

interesaaat:
Closes #613

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ with open(README) as f:
 
 install_requires = [
     "numpy>=1.15",
-    "onnxconverter-common>=1.6.0,<=1.11.0",
+    "onnxconverter-common>=1.6.0",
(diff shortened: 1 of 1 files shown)


ksaur:
Verified that it  pulled `Downloading onnxconverter_common-1.12.1-py2.py3-none-any.whl`.    Let me wipe the cache  and rerun

ksaur (approved):

PR #628, opened 2022-08-31
Corrects some typos

melanciani:
- Fix minor typos in convert method and supported functions.
- No issue is referenced as specified in the guidelines

### DIFF
--- a/hummingbird/ml/convert.py
+++ b/hummingbird/ml/convert.py
@@ -411,7 +411,7 @@ def convert(model, backend, test_input=None, device="cpu", extra_config={}):
     *Convert* supports [Sklearn], [LightGBM], [XGBoost], [ONNX], and [SparkML] models.
     For *LightGBM* and *XGBoost* currently only the Sklearn API is supported.
     The detailed list of models and backends can be found at `hummingbird.ml.supported`.
-    The *onnx* backend requires either a test_input of a the initial types set through the exta_config parameter.
+    The *onnx* backend works best with a test_input, but we try to generate one if none is provided.
     The *torch.jit* and *tvm* backends require a test_input.
     For *tvm* backend, the output container can do prediction only on the test data with the same size as test_input.
     [Sklearn]: https://scikit-learn.org/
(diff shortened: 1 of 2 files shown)


melanciani:
I'm however not sure about the ONNX parameter sentence : `The *onnx* backend requires a test_input of the initial types, set through the extra_config parameter.` Shouldn't the `set through the extra_config parameter` be removed ?

ksaur:
Hi @RomanBredehoft Thanks for catching these!

> I'm however not sure about the ONNX parameter sentence : `The *onnx* backend requires a test_input of the initial types, set through the extra_config parameter.` Shouldn't the `set through the extra_config parameter` be removed ?

You're right, this is not set through `extra_config`.  I know that things work best with ONNX when the user provides initial input  (ex: `convert(onnx_ml_model, "onnx", Xtest)`), but if they don't we try to [generate it](
https://github.com/microsoft/hummingbird/blob/main/hummingbird/ml/convert.py#L168)...not always successfully!

I think we should change this to something like:
`The *onnx* backend works best with a test_input, but we try to generate one if none is provided.`  Thoughts @interesaaat ?

interesaaat:
> Hi @RomanBredehoft Thanks for catching these!
> 
> > I'm however not sure about the ONNX parameter sentence : `The *onnx* backend requires a test_input of the initial types, set through the extra_config parameter.` Shouldn't the `set through the extra_config parameter` be removed ?
> 
> You're right, this is not set through `extra_config`. I know that things work best with ONNX when the user provides initial input (ex: `convert(onnx_ml_model, "onnx", Xtest)`), but if they don't we try to [generate it](https://github.com/microsoft/hummingbird/blob/main/hummingbird/ml/convert.py#L168)...not always successfully!
> 
> I think we should change this to something like: `The *onnx* backend works best with a test_input, but we try to generate one if none is provided.` Thoughts @interesaaat ?

Yes, agreed.

ksaur:
Hi @RomanBredehoft , can you make that change on line 414? (`The *onnx* backend works best with a test_input, but we try to generate one if none is provided.`)

melanciani:
@ksaur done ! just rebased and made the change

interesaaat (approved):

interesaaat:
Thanks for the contribution!

PR #630, opened 2022-08-31
Fix linear models conversion when fit_intercept set to False

### LINKED ISSUE #629: Conversion of linear models fails when fit_intercept set to False
Hi! 

I've encountered a conversion error when trying to convert linear models when fit_intercept is set to False ! A PR will follow as I believe I have found the issue

Code to reproduce : 
```
from sklearn.linear_model import LinearRegression
from sklearn.datasets import make_regression

x, y = make_regression()

model = LinearRegression(fit_intercept=False)
model.fit(x, y)

torch_model = convert(model, backend="torch")
```

Error : 
`AttributeError: 'float' object has no attribute 'reshape'`

melanciani:
closes #629


ksaur (approved):
Thanks for this bugfix!

PR #631, opened 2022-09-07
Added support for more decision conditions in trees and ONNX conversion

grafail:
Resolve issue mentioned in #226

### DIFF
--- a/tests/test_onnxml_decision_tree_converter.py
+++ b/tests/test_onnxml_decision_tree_converter.py
@@ -4,22 +4,21 @@ Tests lightgbm->onnxmltools->hb conversion for lightgbm models.
 import unittest
 import warnings
 
-import sys
-import os
-import pickle
 import numpy as np
 from sklearn.tree import DecisionTreeClassifier, DecisionTreeRegressor
 from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
 from onnxconverter_common.data_types import FloatTensorType
 
 from hummingbird.ml import convert
-from hummingbird.ml import constants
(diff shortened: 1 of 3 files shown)


ksaur:
Thank you so much for this contribution!  Can you please add a test case that shows this? Maybe similar to the one shown in the issue?  Thanks!

grafail:
I verified that the example in #226 still triggers the issue (by using BRANCH_LT) and added a test for it.

tests/test_onnxml_decision_tree_converter.py
```diff
@@ -252,6 +255,24 @@ def test_random_forest_multi_random(self):
...
+        n_total = 100
+        np.random.seed(0)
+        X = np.random.rand(n_total, n_features)
+        X = np.array(X, dtype=np.float32)
+        y = np.random.randint(2, size=n_total)
+
+        # Create XGBoost Model
+        model = xgb.XGBClassifier()
+        model.fit(X, y)
+        self._test_classifier(X, model, conversion_function=convert_xgboost)
```
interesaaat: Is this actually required? I think that if we call `convert` directly, internally it should pick `convert_sklearn` or `convert_xgboost` automatically based on the model type.

grafail: I think it is required, as the issue can only be recreated when converting via onnx-ml. But indeed it seems `convert_model` from `onnxmltools` can be used to auto select the right converter.

interesaaat (approved):

interesaaat:
Thanks for the contribution!

PR #655, opened 2022-11-25
chore: update deprecated calls

stillmatic:
Addresses a few deprecation warnings which I saw:

- Deprecated SKLearn API calls
- Replaces `distutils.version` with `packaging.version` calls. Calls `parse` if the version string is package provided, calls `Version` if the version string is hard coded.
- Drops the deprecated`distutils` entirely to use `setuptools` for everything.
- Drops unnecessary Python 2 backports.

I have signed the Microsoft CLA.

### DIFF
--- a/benchmarks/datasets.py
+++ b/benchmarks/datasets.py
@@ -23,7 +23,6 @@
 # Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
 
 import os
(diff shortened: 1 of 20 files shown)


interesaaat (approved):

interesaaat:
Awesome, thanks for doing this!

PR #674, opened 2023-01-10
allow for identity POST_TRANSFORM

jppgks:
In https://github.com/ludwig-ai/ludwig, we require access to the raw logits since we already have our own "post-transform" logic. With this PR, a user can predict raw logits by using the following `extra_config`:

```python
from hummingbird.ml.operator_converters import constants

lgbm_model = ...
convert(lgbm_model, "torch", extra_config={constants.POST_TRANSFORM: None})
```

### DIFF
--- a/hummingbird/ml/operator_converters/_gbdt_commons.py
+++ b/hummingbird/ml/operator_converters/_gbdt_commons.py
@@ -13,6 +13,7 @@ import numpy as np
 from . import constants
 from ._tree_commons import get_tree_params_and_type, get_parameters_for_tree_trav_common, get_parameters_for_gemm_common
 from ._tree_commons import (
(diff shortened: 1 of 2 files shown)


jppgks:
@ksaur would you have bandwidth to review this small PR? Many thanks!

ksaur (changes requested):
Thanks for the contribution! 
Can you please add a test case that demonstrates use?

jppgks:
@microsoft-github-policy-service agree company="Predibase, Inc."

jppgks:
@ksaur Thanks for reviewing! I just added a test case and agreed to the CLA.

ksaur (approved):
thanks!

Issue #699, opened 2023-05-08
New onnx - failed workflow run


ksaur:
From our [Friday runner](https://github.com/microsoft/hummingbird/actions/runs/4897649694), something is wrong with the `skl2onnx` install

```
        if not utils.skl2onnx_installed():
>           raise RuntimeError('skl2onnx is not installed. Please install skl2onnx to use this feature.')
E           RuntimeError: skl2onnx is not installed. Please install skl2onnx to use this feature.
```

ksaur:
`skl2onnx` is getting installed correctly, but there is a problem with it that makes it unusable, and so the tests fail saying that it is uninstalled.

I this there is a [problem with onnx 1.14](https://github.com/onnx/onnx/issues/5202)


```
root@a033bfa45751:/# python3
Python 3.8.10 (default, Mar 13 2023, 10:26:41)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import skl2onnx
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/skl2onnx/__init__.py", line 15, in <module>
    from .convert import convert_sklearn, to_onnx, wrap_as_onnx_mixin  # noqa
  File "/usr/local/lib/python3.8/dist-packages/skl2onnx/convert.py", line 6, in <module>
    from .common._topology import convert_topology
  File "/usr/local/lib/python3.8/dist-packages/skl2onnx/common/_topology.py", line 29, in <module>
    from ..proto.onnx_helper_modified import (
  File "/usr/local/lib/python3.8/dist-packages/skl2onnx/proto/onnx_helper_modified.py", line 11, in <module>
    from onnx.helper import (  # noqa
ImportError: cannot import name '_to_bytes_or_false' from 'onnx.helper' (/usr/local/lib/python3.8/dist-packages/onnx/helper.py)
```

ksaur:
Will leave this issue open until we can remove the pin

PR #700, opened 2023-05-08
Problem with ONNX=1.14 and skl2onnx

ksaur:
Addresses issue #699 where ONNX imports clashes causes bug that causes import to fail (but install to succeed)


ksaur:
Will unpin after https://github.com/onnx/sklearn-onnx/pull/988 is in the release; blocked for now

interesaaat (approved):

PR #705, opened 2023-05-15
Pinning holidays

### LINKED ISSUE #704: Compilation issue in Prophet
The build in main currently fails with the error `TypeError: object.__init__() takes exactly one argument (the instance to initialize)`

interesaaat:
Momentarily fixes #704


ksaur:
Thanks for the quick fix! I'll leave the issue open for now so we can sort it out later

ksaur (approved):

PR #643, opened 2022-10-25
Add support for Python 3.11

mshr-h:
Waiting for the update
- [x] PyTorch for Windows/macOS
  - https://github.com/pytorch/pytorch/issues/86566
  - We may have to wait 1.14 release https://github.com/pytorch/pytorch/issues/86566#issuecomment-1315946802
  - #691 
- [x] ONNX 1.13 will support Python 3.11
  - [Logistics for ONNX Release 1.13.0 · onnx/onnx Wiki](https://github.com/onnx/onnx/wiki/Logistics-for-ONNX-Release-1.13.0)
  - Released https://github.com/onnx/onnx/releases/tag/v1.13.0
- [x] ONNX Runtime
  - https://github.com/microsoft/onnxruntime/issues/14880#issuecomment-1452037857
  - ORT with Python 3.11 support will be released around June/July.
- [ ] Update TVM.
- [ ] Solve Prophet test failure on MacOS with Python 3.11
  - https://github.com/microsoft/hummingbird/actions/runs/5132441794/jobs/9233734808?pr=643

PR #709, opened 2023-06-01
Update apache-tvm to v0.15.0

mshr-h:
TVM doesn't provide a stable version of a binary package for Python 3.11. So we'll build it from the source in the CI pipeline.
Also, we'll update TVM to the latest stable release.
Related to #643

TODO
- [x] run pipeline again after the TVM v0.15.0 release
- [x] cleanup pipeline

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -26,7 +26,9 @@ jobs:
         os: [ubuntu-22.04, macos-12, windows-2022]
         python-version: ['3.8', '3.9', '3.10', '3.11']
     env:
+      TVM_VERSION_TAG: v0.15.0
       PYTORCH_VERSION: 2.0.0
+      LLVM_VERSION: 14
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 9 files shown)


mshr-h:
TVM v0.12.0's pytorch frontend failes to import HB's pytorch model in v1.12.0.
But it's ok when use v0.11.1

mshr-h:
I did git bisect and found that the below commit has a problem.
[[Relay][Frontend] Span Filling PyTorch (#14050) · apache/tvm@e9cf04e](https://github.com/apache/tvm/commit/e9cf04e0e4ec325ca665fe2d262b481985c8cf42)

mshr-h:
Seems like there's a bug in the PyTorch frontend. Opened new issue https://github.com/apache/tvm/issues/16150.

ksaur:
Was that last failure transient, or did something new break unrelated to TVM? Please let me know if you are blocked on something and I can try to help!

mshr-h:
The CI test has passed but I don't know why...

mshr-h:
We'll wait for the TVM v0.15.0 which is going to be released on 22 Jan 2024.
https://github.com/apache/tvm/issues/16277
After the release, we'll run the CI again.

mshr-h:
CI is failing again.
When TVM was installed from the source, it was ok. But it was restored from the cache, the `test_extra_trees_tvm_tree_trav_regressor_converter` test failed.

Ok log: https://github.com/microsoft/hummingbird/actions/runs/7692350896/job/20959123441
Failed log: https://github.com/microsoft/hummingbird/actions/runs/7693930724/job/20963638667

mshr-h:
Seems like tests are a bit flaky when TVM is installed from the cache.

Only failed on ubuntu+python3.9: https://github.com/microsoft/hummingbird/actions/runs/7710611530/job/21014292656
Only failed on ubuntu+python3.8: https://github.com/microsoft/hummingbird/actions/runs/7711820622/job/21017938809

@ksaur @interesaaat 
Do you have any ideas on this problem?

ksaur:
> Seems like tests are a bit flaky when TVM is installed from the cache.
> 
> Only failed on ubuntu+python3.9: https://github.com/microsoft/hummingbird/actions/runs/7710611530/job/21014292656 Only failed on ubuntu+python3.8: https://github.com/microsoft/hummingbird/actions/runs/7711820622/job/21017938809
> 
> @ksaur @interesaaat Do you have any ideas on this problem?

Hmmm, `: line 1:  2222 Aborted                 (core dumped) pytest`  since it's transient/"bit flaky", i wonder if it is a memory issue. have you ever seen it on your local machine or could get the coredump?  And it's _only_ when from cache?

interesaaat:
> Seems like tests are a bit flaky when TVM is installed from the cache.
> 
> Only failed on ubuntu+python3.9: https://github.com/microsoft/hummingbird/actions/runs/7710611530/job/21014292656 Only failed on ubuntu+python3.8: https://github.com/microsoft/hummingbird/actions/runs/7711820622/job/21017938809
> 
> @ksaur @interesaaat Do you have any ideas on this problem?

My 2c is that 30 for some reason is too much for `TVM_MAX_FUSE_DEPTH`. Can we try with 10 and see if it passes?

mshr-h:
Thank you for your comments.

@ksaur 
> have you ever seen it on your local machine or could get the coredump?

No.

> And it's only when from cache?

It happens both from the source and from the cache.

@interesaaat 
I tried `TVM_MAX_FUSE_DEPTH: 10` but still flaky :(

1st run (Ok): https://github.com/microsoft/hummingbird/actions/runs/7722097282
2nd run (Ng): https://github.com/microsoft/hummingbird/actions/runs/7723327148

Both runs are tvm from the cache. And always failed only on ubuntu, no fail on macOS.

mshr-h:
According to the GitHub official docs, the GitHub-hosted runners have 16GB RAM for Linux and 14GB RAM for macOS.
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

ksaur:
> 1st run (Ok): https://github.com/microsoft/hummingbird/actions/runs/7722097282 
> 2nd run (Ng): https://github.com/microsoft/hummingbird/actions/runs/7723327148
> 
> Both runs are tvm from the cache. And always failed only on ubuntu, no fail on macOS.

Hmm from the Ng run: 
```
Current thread 0x00007f3036869b80 (most recent call first):
  Garbage-collecting
  File "/home/runner/work/hummingbird/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 215 in __del__
  File "/home/runner/work/hummingbird/tvm/python/tvm/_ffi/_ctypes/packed_func.py", line 119 in _make_tvm_args
  ...
```
I'm not exactly sure what it's up to, but it definitely looks like a memory issue!

> According to the GitHub official docs, the GitHub-hosted runners have 16GB RAM for Linux and 14GB RAM for macOS.

Let me see if we can do [large runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#larger-runners)....

ksaur:
Ok I looked into large runners. For some reason that I cannot understand, we do not have access to Github Enterprise with this repo, and therefore can't setup large runners.  As a backup plan, we could try self-hosted runners in our Azure sub, but that will take a long time to setup and we'd need to figure out a budget.

In the meantime, can you please try something even smaller? Like `TVM_MAX_FUSE_DEPTH: 10` to `1` or something? (not sure what is a sensible value). Or smaller `n_estimators` etc?    (or just put 0 and it won't fuse, that should avoid OOM for sure!)

mshr-h:
I think we have to check the max memory usage during the test on the local machine.

ksaur:
Thanks for pushing so hard on this!! :)  I'm wondering if it's too risky to do something like a [retry](https://stackoverflow.com/questions/71574593/how-to-automatically-retry-github-action-jobs-on-failure) on those tests?  But that could take a long time.

Maybe we could also mark those tests as ok-to-fail (just for the ubuntu TVM ones with the memory issues).  Both of those ideas are a bit sloppy, but I'm not sure of a better way to unblock it.

mshr-h:
pytest-memray provides a marker to limit memory usage on a specific test.
[Usage - pytest-memray](https://pytest-memray.readthedocs.io/en/latest/usage.html#pytest.mark.limit_memory)

Can we do something like if the test try to allocate more than XX MB, finish the execution of the test and mark as **skipped because memory usage are too high**.

mshr-h:
The memory issue is happening in the TVM side.
I'll work on some more days to see if there's any way to solve it.

mshr-h:
> Thanks for pushing so hard on this!! :)  I'm wondering if it's too risky to do something like a [retry](https://stackoverflow.com/questions/71574593/how-to-automatically-retry-github-action-jobs-on-failure) on those tests?  But that could take a long time.
> 
> 
> 
> Maybe we could also mark those tests as ok-to-fail (just for the ubuntu TVM ones with the memory issues).  Both of those ideas are a bit sloppy, but I'm not sure of a better way to unblock it.

Or maybe we can temporary disable all the TVM tests on Ubuntu. They're also tested on macOS.

ksaur:
> > Thanks for pushing so hard on this!! :)  I'm wondering if it's too risky to do something like a [retry](https://stackoverflow.com/questions/71574593/how-to-automatically-retry-github-action-jobs-on-failure) on those tests?  But that could take a long time.
> > Maybe we could also mark those tests as ok-to-fail (just for the ubuntu TVM ones with the memory issues).  Both of those ideas are a bit sloppy, but I'm not sure of a better way to unblock it.
> 
> Or maybe we can temporary disable all the TVM tests on Ubuntu. They're also tested on macOS.

I'm fine with this solution (skipping TVM tests on Ubu)! (Unless there are OS-specific TVM issues)? What do you think @interesaaat?

mshr-h:
Note for the memory issue investigation
- When does it happen?
  - when running pytest with extra dependencies on Ubuntu+python 3.8-3.11 on GitHub Actions.
  - Not happen on macOS or Windows.
  - Sometimes it passes, sometimes it fails.
  - Haven't seen a similar issue on my local Ubuntu machine.
- What's the error message look like?
  - `: line 1:  2222 Aborted                 (core dumped) pytest`
    - Link to a pipeline log: https://github.com/microsoft/hummingbird/actions/runs/7723327148/job/21053180461
    - It seems like the error is happening when the TVM tries to build the module.
- What I tried?
  - smaller estimators (`n_estimators=5`) for `test_random_forest_tvm_tree_trav_classifier_converter`.
  - disabled op fusion for TVM (`TVM_MAX_FUSE_DEPTH: 0`)
  - limited memory usage by systemd e.g. `systemd-run --scope -p MemoryMax=8G --user pytest`
  - linked LLVM statically when TVM build
  - updated PyTorch to 2.2.0

ksaur:
Hi @mshr-h thanks for all your work on this!! (Especially thank you for the nice summary of the memory issue.) Is the PR still a WIP?  I like the `is_on_github_actions` :)  Please let us know if you get stuck or want us to take a look!!

mshr-h:
Should be good.

@ksaur @interesaaat 
Thank you for your support! Can you review it?

ksaur:
This looks good to me!! @interesaaat take a look and merge if ready

ksaur (approved):
Thank you so much for all your hard work! 🎊

PR #713, opened 2023-06-16
Adding hash to save/load

ksaur:


hummingbird/ml/_utils.py
```diff
@@ -237,7 +237,7 @@ def from_strings_to_ints(input, max_string_length):
     return np.array(input, dtype="|S" + str(max_string_length)).view(np.int32).reshape(shape)
 
 
-def load(location):
+def load(location, digest=None):
```
ksaur, line 240: TODO: Add pydoc strings on this function

interesaaat (approved):
I am not sure on whether we should add the option to provide the shared key but let's go with this for the moment

Issue #715, opened 2023-06-19
TVM MacOS pipeline intermittent failures


ksaur:
Recently we've been seeing failures like [this](https://github.com/microsoft/hummingbird/actions/runs/5316237754/jobs/9625542616?pr=713#step:19:3730) and [this](https://github.com/microsoft/hummingbird/actions/runs/5295167134/jobs/9585215143#step:19:3734):

```
Processing ./dist/tvm-0.10.0-cp38-cp38-macosx_11_0_x86_64.whl
ERROR: Cannot install tvm 0.10.0 (from /Users/runner/work/hummingbird/tvm/python/dist/tvm-0.10.0-cp38-cp38-macosx_10_15_x86_64.whl) and tvm 0.10.0 (from /Users/runner/work/hummingbird/tvm/python/dist/tvm-0.10.0-cp38-cp38-macosx_11_0_x86_64.whl) because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts

The conflict is caused by:
    The user requested tvm 0.10.0 (from /Users/runner/work/hummingbird/tvm/python/dist/tvm-0.10.0-cp38-cp38-macosx_10_15_x86_64.whl)
    The user requested tvm 0.10.0 (from /Users/runner/work/hummingbird/tvm/python/dist/tvm-0.10.0-cp38-cp38-macosx_11_0_x86_64.whl)

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

This seems to happen every other time, todo: investigate. 
cc: @mshr-h , have you seen this other places?

mshr-h:
No, I haven't seen any similar message.

mshr-h:
@ksaur 
Can you try to remove the GHA cache and run the pipeline again?

I ran again the pipeline and found that there are multiple whl package in python/dist.
The 10_15 whl was created in the previous pipeline execution and restored from the cache.
Recent pipeline executions create 11_0 whl, then try to install them at the same time which caused the dependency conflict.

> -rw-r--r--  1 runner  staff  42055067 Feb 24 22:03 tvm-0.10.0-cp38-cp38-macosx_10_15_x86_64.whl
[...]

PR #716, opened 2023-06-20
[WIP]Investigate #715

mshr-h:
#715

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -126,6 +126,7 @@ jobs:
       run: |
         python -m pip install -U wheel packaging
         python setup.py bdist_wheel
+        ls -l dist/
         python -m pip install dist/tvm-*.whl
 
     # We don't run pytest for Linux py3.8 since we do coverage for that case.


mshr-h:
> -rw-r--r--  1 runner  staff  42055067 Feb 24 22:03 tvm-0.10.0-cp38-cp38-macosx_10_15_x86_64.whl
-rw-r--r--  1 runner  staff  42055470 Jun 20 02:39 tvm-0.10.0-cp38-cp38-macosx_11_0_x86_64.whl

There are multiple whl  packages in python/dist/ and try to install them simultaneously.
One of them was restored from the cache.
I think we need to exclude python/dist/ from the cache. (Just remove the directory before caching?)

ksaur:
Clearing the cache seems to work, thanks so much for that suggestion!! But I see what you mean, do we also need to exclude the `dist` from the dir before caching?

ksaur:
<del>Now some new [problem](https://github.com/microsoft/hummingbird/actions/runs/5339192071/jobs/9678881224) this time with TVM+pypi; I'm hoping intermittent (will try again in an hour, seems an outage)</del>
Was transient, yay

mshr-h:
Maybe we can avoid the issue by moving the wheel building command to the `Build TVM if Mac` step.

Issue #727, opened 2023-07-18
TVM + Mac


ksaur:
Another issue in the [pipeline](https://github.com/microsoft/hummingbird/actions/runs/5592796893/jobs/10226035572):

```
Compiling tvm/_ffi/_cython/core.pyx because it changed.
Error compiling Cython file:
------------------------------------------------------------
...
        if self.c_is_view != 0:
            raise ValueError("to_dlpack do not work with memory views")
        with nogil:
            c_api_ret_code = TVMArrayToDLPack(self.chandle, &dltensor)
        CHECK_CALL(c_api_ret_code)
        return pycapsule.PyCapsule_New(dltensor, _c_str_dltensor, _c_dlpack_deleter)
                                                                  ^
------------------------------------------------------------

tvm/_ffi/_cython/./ndarray.pxi:116:66: Cannot assign type 'void (object) except *' to 'PyCapsule_Destructor'
[1/1] Cythonizing tvm/_ffi/_cython/core.pyx
Traceback (most recent call last):
  File "setup.py", line 238, in <module>
    ext_modules=config_cython(),
  File "setup.py", line 154, in config_cython
    return cythonize(ret, compiler_directives={"language_level": 3})
  File "/Users/runner/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/site-packages/Cython/Build/Dependencies.py", line 1[13](https://github.com/microsoft/hummingbird/actions/runs/5592796893/jobs/10226035572#step:20:14)4, in cythonize
    cythonize_one(*args)
  File "/Users/runner/hostedtoolcache/Python/3.8.[17](https://github.com/microsoft/hummingbird/actions/runs/5592796893/jobs/10226035572#step:20:18)/x64/lib/python3.8/site-packages/Cython/Build/Dependencies.py", line 1[30](https://github.com/microsoft/hummingbird/actions/runs/5592796893/jobs/10226035572#step:20:31)1, in cythonize_one
    raise CompileError(None, pyx_file)
Cython.Compiler.Errors.CompileError: tvm/_ffi/_cython/core.pyx
Error: Process completed with exit code 1.
```

I don't want to block all of Hummingbird if TVM install fails on MacOS.
Going forward, I think we need to remove MacOS+TVM compile from failout out the entire pipeline, else too many potential users will be blocked on contributing.

ksaur:
[...]

PR #728, opened 2023-07-19
continue-on-error: true for mac tvm install fail

ksaur:
For https://github.com/microsoft/hummingbird/issues/727

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -100,6 +100,7 @@ jobs:
       if: ${{ startsWith(matrix.os, 'macos') }}
       run: |
         brew install llvm@14
+      continue-on-error: true
     - name: Fetch and prepare TVM for compilation if Mac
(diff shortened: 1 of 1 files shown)


ksaur:
This does work, but now it fails silently.  (The pipeline continues, runs all non-TVM tests on mac, and gives green checkmark without showing any warnings/etc.) This seems not a perfect solution either. There's probably some cleaner way to get alerts on this while not blocking everything. 

Thoughts Matteo/Masahiro?

Screenshot 2023-07-19 at 11 07 57 AM

mshr-h (approved):
Looks good to me.

interesaaat:
Is this a draft or good to go?

ksaur:
I just wanted both of your inputs to see if this is behavior we wanted. It may make things easier in future, but I also want to make sure mac TVM is supported when/where possible!

ksaur:
Leaving this as a draft....ready to go if needed later (hopefully not). :)

Issue #736, opened 2023-09-22
onnxruntime==1.16.0 release breaks tests


ksaur:
onnxruntime==1.16.0 was [released](https://github.com/microsoft/onnxruntime/releases/tag/v1.16.0)
and [tests](https://github.com/microsoft/hummingbird/actions/runs/6279215738/job/17054486386) fail.

Pinning for now. #737

ksaur:
For the most part, we need to fix:
```
>           raise ValueError(
                f"This ORT build has {available_providers} enabled. "
                "Since ORT 1.9, you are required to explicitly set "
                "the providers parameter when instantiating InferenceSession. For example, "
                f"onnxruntime.InferenceSession(..., providers={available_providers}, ...)"
```

ksaur:
https://github.com/microsoft/onnxruntime/issues/17631  Apparently they are working on a fix

ksaur:
closed with #738 and https://github.com/microsoft/onnxruntime/commit/1c245e6775201fec7a1d269a71deb0f3af6f7bea . Thanks all!

PR #738, opened 2023-10-13
Unpinning onnxruntime==0.15.1

mshr-h:
#736 
#737 
Seems like the issue was fixed on 0.16.1.


mshr-h:
This is the fix commit.
https://github.com/microsoft/onnxruntime/commit/1c245e6775201fec7a1d269a71deb0f3af6f7bea

ksaur (approved):
Thanks for the follow-up!

Issue #744, opened 2023-12-15
ONNXMLGather breaks tests


ksaur:
[failed runner](https://github.com/microsoft/hummingbird/actions/runs/7227155960/job/19694227010)

todo investigate: `ValueError: Unsupported conversion for operator ONNXMLGather`

```
=========================== short test summary info ===========================
FAILED tests/test_backends.py::TestBackends::test_onnx_no_test_data_int - ValueError: Unsupported conversion for operator ONNXMLGather
FAILED tests/test_onnxml_one_hot_encoder_converter.py::TestONNXOneHotEncoder::test_model_one_hot_encoder_string - ValueError: Unsupported conversion for operator ONNXMLGather
FAILED tests/test_onnxml_one_hot_encoder_converter.py::TestONNXOneHotEncoder::test_one_hot_encoder_onnx2 - ValueError: Unsupported conversion for operator ONNXMLGather
FAILED tests/test_onnxml_one_hot_encoder_converter.py::TestONNXOneHotEncoder::test_one_hot_encoder_onnx_int - ValueError: Unsupported conversion for operator ONNXMLGather
FAILED tests/test_onnxml_one_hot_encoder_converter.py::TestONNXOneHotEncoder::test_one_hot_encoder_onnx_int64 - ValueError: Unsupported conversion for operator ONNXMLGather
==== 5 failed, 530 passed, 105 skipped, 223 warnings in 309.38s (0:05:09) =====

```

ksaur:
[Old](https://github.com/microsoft/hummingbird/actions/runs/7079925054/job/19267128674):
```
Collecting skl2onnx>=1.7.0
  Downloading skl2onnx-1.15.0-py2.py3-none-any.whl.metadata (2.3 kB)
```

[New](https://github.com/microsoft/hummingbird/actions/runs/7227155960/job/19694227010):
```
Collecting skl2onnx>=1.7.0
  Downloading skl2onnx-1.16.0-py2.py3-none-any.whl.metadata (3.2 kB)

```

ksaur:
It appears that `ArrayFeatureAbstractor` was [changed](https://github.com/onnx/sklearn-onnx/compare/1.15.0...1.16.0#diff-302f1522fefcd7edaddb5cc7dfbee7ae4d2368310b4652ead5e8d79fd27a3788L104) to `Gather` in https://github.com/onnx/sklearn-onnx/commit/262abfc78b0079d1f2eb8113bfbf67abd50b1bd3

PR #745, opened 2023-12-25
Unpinning xgboost version

### LINKED ISSUE #732: XGBoost 2.0.0 breaks tests
XGB 2.0.0 was [released](https://github.com/dmlc/xgboost/releases/tag/v2.0.0) last week (Sept 12, 2023).  This causes 19 tests to [fail](https://github.com/microsoft/hummingbird/actions/runs/6202924126/job/16842625679). 

Many errors are related to NDCG ("NDCG is now the default objective function.")

mshr-h:
Fix #732
Seems like [1.12.0](https://github.com/onnx/onnxmltools/releases/tag/1.12.0) has the fix?

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -22,6 +22,7 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-22.04, macos-12, windows-2022]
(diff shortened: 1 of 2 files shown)


ksaur:
Thanks! It should, but we need to first fix #744.  My goal is to patch by end of week!

ksaur:
please rebase and hopefully it works! :)

PR #764, opened 2024-02-28
fixing XGBoostRegression issue with base_predicitions

giriprasad51:
FAILED tests/test_xgboost_converter.py::TestXGBoostConverter::test_float64_xgb_ranker_converter  - AssertionError:
FAILED tests/test_xgboost_converter.py::TestXGBoostConverter::test_float64_xgb_regressor_converter - AssertionError:
FAILED tests/test_xgboost_converter.py::TestXGBoostConverter::test_run_xgb_pandas - AssertionError:

Above test cases are passed

### DIFF
--- a/hummingbird/ml/operator_converters/xgb.py
+++ b/hummingbird/ml/operator_converters/xgb.py
@@ -135,7 +135,7 @@ def convert_sklearn_xgb_regressor(operator, device, extra_config):
         feature_names = operator.raw_operator.get_booster().feature_names
         if feature_names is not None:
             extra_config[constants.FEATURE_NAMES] = feature_names
-    base_prediction =  float(eval(operator.raw_operator.get_booster().save_config().replace('false', "False").replace("true", "True"))["learner"]["learner_model_param"]["base_score"])
+    base_prediction = float(eval(operator.raw_operator.get_booster().save_config().replace('false', "False").replace("true", "True"))["learner"]["learner_model_param"]["base_score"])
     if base_prediction is None:
         base_prediction = [0.5]
     if type(base_prediction) is float:
--- a/hummingbird/ml/operator_converters/xgb.py
+++ b/hummingbird/ml/operator_converters/xgb.py
@@ -136,11 +136,9 @@ def convert_sklearn_xgb_regressor(operator, device, extra_config):
         if feature_names is not None:
             extra_config[constants.FEATURE_NAMES] = feature_names
     import json
-    def get_basescore(model) -> float:
+    def get_basescore(model):
         """Get base score from an XGBoost sklearn estimator."""
-        base_score = float(
-            json.loads(model.get_booster().save_config())["learner"]["learner_model_param"]["base_score"]
-        )
+        base_score = float(json.loads(model.get_booster().save_config())["learner"]["learner_model_param"]["base_score"])
         return base_score
     base_prediction = get_basescore(operator.raw_operator)
     if base_prediction is None:
--- a/hummingbird/ml/operator_converters/xgb.py
+++ b/hummingbird/ml/operator_converters/xgb.py
@@ -135,15 +135,7 @@ def convert_sklearn_xgb_regressor(operator, device, extra_config):
         feature_names = operator.raw_operator.get_booster().feature_names
         if feature_names is not None:
             extra_config[constants.FEATURE_NAMES] = feature_names
-    import json
-
-    def get_basescore(model):
-
-        """Get base score from an XGBoost sklearn estimator."""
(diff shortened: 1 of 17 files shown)


giriprasad51:
Hii @ksaur I made some changes in the code please go through it. Thank you!

ksaur:
Hello! Was this failing somewhere? Is this related to an open issue? Can you please show me where the assertion errors you posted were coming from?  (what version of xgb, etc).  I'm only aware of the #732 but that was with xgb==2.0  thanks!

giriprasad51:
Hii @ksaur  xgb = 2.0.3
By running test cases
!pytest /content/hummingbird/tests

From  this 
=============== 18 failed, 443 passed, 203 skipped, 65 warnings in 164.71s (0:02:44) ===============

To

=============== 15 failed, 446 passed, 203 skipped, 65 warnings in 167.01s (0:02:47) ===============

You can reffer this colab notebook : 
 https://colab.research.google.com/drive/1hUqyl2LafAMnN01JGju5Huzv8JRQeQhI

giriprasad51:
https://github.com/microsoft/hummingbird/blob/3eb83937d3a3776b4b3eb82c1011243833a4bcd3/hummingbird/ml/operator_converters/xgb.py#L138

Even though xgboost regressor contains base score here it's getting None

ksaur:
cc: @interesaaat to check it out (since he's worked with this section of code quite a bit)

hummingbird/ml/operator_converters/xgb.py
```diff
@@ -135,7 +135,15 @@ def convert_sklearn_xgb_regressor(operator, device, extra_config):
         feature_names = operator.raw_operator.get_booster().feature_names
         if feature_names is not None:
             extra_config[constants.FEATURE_NAMES] = feature_names
-    base_prediction = operator.raw_operator.base_score
+    import json
+
+    def get_basescore(model):
+
+        """Get base score from an XGBoost sklearn estimator."""
+        base_score = float(json.loads(model.get_booster().save_config())["learner"]["learner_model_param"]["base_score"])
```
interesaaat: Can you do the same without having to dump the model into json and then load it again? Like can you access the `base_score` from `operator.raw_operator.lerner.lernel_model_parm`?

interesaaat (dismissed):

giriprasad51:
Hii @interesaaat  As per your suggestion i modify the code please go through it

interesaaat:
> Hii @interesaaat As per your suggestion i modify the code please go through it

That is not what I meant because you are still saving the model. I think there is a way to access the `base_score` directly without having to save the model.

mshr-h:
I think the only way to access the model parameter is by using the `save_config()` method.
I grepped the XGBoost source code and found this utility function just for testing.
https://github.com/dmlc/xgboost/blob/v2.0.3/python-package/xgboost/testing/updater.py#L12

It's quite similar what @giriprasad51 is doing.
We can use `get_basescore()` by `from xgboost.testing.updater import get_basescore` but it's only available from 2.0.0.

ksaur:
> I think the only way to access the model parameter is by using the `save_config()` method. I grepped the XGBoost source code and found this utility function just for testing. https://github.com/dmlc/xgboost/blob/v2.0.3/python-package/xgboost/testing/updater.py#L12
> 
> It's quite similar what @giriprasad51 is doing. We can use `get_basescore()` by `from xgboost.testing.updater import get_basescore` but it's only available from 2.0.0.

Thanks @mshr-h !   Hmm Ok, what do you think @interesaaat ? Should we try to do this just here, or wait and have a higher level solution that's more of a broad xgb==2.0 fix? I am open to either!

interesaaat:
> > I think the only way to access the model parameter is by using the `save_config()` method. I grepped the XGBoost source code and found this utility function just for testing. https://github.com/dmlc/xgboost/blob/v2.0.3/python-package/xgboost/testing/updater.py#L12
> > It's quite similar what @giriprasad51 is doing. We can use `get_basescore()` by `from xgboost.testing.updater import get_basescore` but it's only available from 2.0.0.
> 
> Thanks @mshr-h ! Hmm Ok, what do you think @interesaaat ? Should we try to do this just here, or wait and have a higher level solution that's more of a broad xgb==2.0 fix? I am open to either!

Interesting. I find it weird that to access a parameter you need to save the model. Like the parameter should be stored somewhere, right? But I haven't looked into the details, so maybe this is the only way? Since it's at conversion this is probably fine, even if not ideal.

ksaur:
Merging this for now.  Maybe it will be steps toward a broader solution for xgb==2.0 in the future!

ksaur (approved):

Issue #766, opened 2024-03-08
Need to update setup.py description


ksaur:
In our setup.py file, we say we are using markdown:

```
-    description="Convert trained traditional machine learning models into tensor computations",
-    long_description=long_description,
-    long_description_content_type="text/markdown",
```

Pypi was not loving this.

```
Uploading hummingbird_ml-0.4.11-py2.py3-none-any.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 174.8/174.8 kB • 00:00 • 1.1 MB/s
INFO     Response from https://upload.pypi.org/legacy/:
         400 The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
    <html>
          <head>
           <title>400 The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.</title>
          </head>
          <body>
           <h1>400 The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.</h1>
           The server could not comply with the request since it is either malformed or otherwise incorrect.<br/><br/>
         The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.


          </body>
         </html>
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
         The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
```

For now just deleted these 3 lines, but we should change the format away from markdown.   See [docs](https://pypi.org/help/#description-content-type)

PR #767, opened 2024-03-09
feat: removed long_description info from setup.py

kachida:
Hi @ksaur , can you please review the below PR#

https://github.com/microsoft/hummingbird/issues/766

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ packages = find_packages()
 assert packages
 
 # read version from the package file.
-with (open(os.path.join(this, "hummingbird/__init__.py"), "r")) as f:
+with open(os.path.join(this, "hummingbird/__init__.py"), "r") as f:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -44,9 +44,6 @@
 setup(
     name="hummingbird-ml",
     version=version_str,
-    description="Convert trained traditional machine learning models into tensor computations",
-    long_description=long_description,
```
ksaur, line 48: Hello thanks for your PR!  The goal is not to delete this text entirely, but to make it legal with the new-style setup.py files.  Something like this:

Screenshot 2024-03-11 at 10 01 18 AM

ksaur (changes requested):

setup.py
ksaur: Please see my comment in the [issue](https://github.com/microsoft/hummingbird/issues/766) "_For now_ just deleted these 3 lines, but we _should change the format_ away from markdown. See [docs](https://pypi.org/help/#description-content-type)" ...   Thanks!!

PR #775, opened 2024-05-30
Deprecating Python 3.8

ksaur:
SKL 1.5.0 no longer supports python 3.8.  Official security support of Python 3.8 will end in October2024.


interesaaat (approved):

ksaur:
@mshr-h do you have any concerns about this? making sure you don't need that version.  If you do, please let me know and we'll add it back!

PR #781, opened 2024-06-27
[CI] Add Python 3.12

mshr-h:


mshr-h:
https://github.com/microsoft/hummingbird/actions/runs/9695338444/job/26754887915?pr=781
The above CI error occurs because PySpark removed its dependency on the distutils package, which was removed in Python 3.12. But the latest PySpark (3.5.1) doesn't include it.
Seems like 4.0.0 will have it. The preview was released on June 3, so I guess the stable version will be released within a few weeks.
https://issues.apache.org/jira/browse/SPARK-45390
#793

segatrade:
Thank you for bringing 3.12 support

Issue #788, opened 2024-08-02
Pipelinefail on ONNX_ML


ksaur:
There was a [release](https://github.com/onnx/onnx/releases/tag/v1.16.2
) yesterday `onnx-1.16.1` to `onnx-1.16.2`

From [runner](https://github.com/microsoft/hummingbird/actions/runs/10222206038/job/28286331182) we have about 50 of these:

```
____________ ERROR collecting tests/test_sklearn_feature_union.py _____________
ImportError while importing test module 'D:\a\hummingbird\hummingbird\tests\test_sklearn_feature_union.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\importlib\__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests\test_sklearn_feature_union.py:15: in <module>
    import hummingbird.ml
hummingbird\ml\__init__.py:22: in <module>
    from .convert import convert, convert_batch  # noqa: F401, E402
hummingbird\ml\convert.py:14: in <module>
    from .operator_converters import constants
hummingbird\ml\operator_converters\__init__.py:20: in <module>
    from .onnx import onnx_operator  # noqa: E402
hummingbird\ml\operator_converters\onnx\onnx_operator.py:12: in <module>
    from onnxconverter_common.registration import register_converter
C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\onnxconverter_common\__init__.py:18: in <module>
    from .data_types import *  # noqa F403
C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\onnxconverter_common\data_types.py:7: in <module>
    import onnx
C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\site-packages\onnx\__init__.py:77: in <module>
    from onnx.onnx_cpp2py_export import ONNX_ML
E   ImportError: DLL load failed while importing onnx_cpp2py_export: A dynamic link library (DLL) initialization routine failed.
```

ksaur:
maybe from this https://github.com/onnx/onnx/issues/6267

PR #789, opened 2024-08-02
mitigating onnx install issue

ksaur:
related to #788 (stop gap, does not close)

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -138,18 +138,18 @@ jobs:
         cd ~/work/hummingbird/tvm/python
         python3 -m pip install -e .
 
-    # We don't run pytest for macos py3.9 since we do coverage for that case.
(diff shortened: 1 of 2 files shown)


interesaaat (approved):

.github/workflows/pythonapp.yml
```diff
@@ -138,19 +138,19 @@ jobs:
         cd ~/work/hummingbird/tvm/python
         python3 -m pip install -e .
 
-    # We don't run pytest for macos py3.9 since we do coverage for that case.
+    # We don't run pytest for ubuntu py3.9 since we do coverage for that case.
```
ksaur, line 141: Now there is some problem with macos + codecov that is blocking the pipeline.  Changing this to unblock, and also since mac is slowest it makes no sense to make it worse! Will fix additional issues in separate PR.

PR #792, opened 2024-08-13
Updated setup.py file

LuniaKunal:
This is my first contribution @ksaur. Tried to Fix issue #766. Sorry for new Pull request.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,9 +45,9 @@ extra_requires = [
 setup(
     name="hummingbird-ml",
     version=version_str,
-    description="Convert trained traditional machine learning models into tensor computations",
-    long_description=long_description,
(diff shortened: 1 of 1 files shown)


ksaur:
Hi thanks for the PR!  The goal was not to remove the lines, but to update them. See this [similar pr](https://github.com/microsoft/hummingbird/pull/767#discussion_r1520087403). 

To push the last release, i just deleted those three lines you deleted before I pushed to pypi, and the end result is that the [description is blank](https://pypi.org/project/hummingbird-ml/0.4.11/#description).  We want to fix it like in the link above where the description is NOT blank. Thanks!

ksaur (changes requested):
please update the syntax of the description rather than just deleting it, thanks!

Issue #235, opened 2020-08-14
Add notebook example for torchscript example


interesaaat:
With #225 we added support for torchscript. It would be good to have a notebook showcasing this new feature.

PR #794, opened 2024-08-29
#235 Added Notebook Example for Torchscript

Shashank1202:


ksaur:
Hello welcome @Shashank1202 !  Thanks for your PR!

tests/torchscript_example_test.py
```diff
@@ -0,0 +1,61 @@
+import torch
```
ksaur, line 1: Please rename the file to start with `test_` to match the others in the folder

tests/torchscript_example_test.py
```diff
@@ -0,0 +1,61 @@
...
+import unittest
+import numpy as np
+
+class TestTorchScriptModel(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        # Load the TorchScript model
+        cls.model = torch.jit.load('notebooks/simple_model_scripted.pt')
```
ksaur, line 10: We like to avoid checking in the model binaries. Can you please recreate it here and use it for your test, and then remove the binary file?

ksaur (changes requested):

